PR #1, opened 2015-06-12
added driver and backend function to measure 2D power

nickhand:
function measure2Dpower returns k, mu, P, N, edges where
edges is [kedges, muedges]. N is the number of modes in each bin


rainwoodman:
In the future we may need to do mu in (-1, 1) e.g. halo matter cross correlation in RSD space?

PR #23, opened 2015-07-07
density profile calculator

energy-berkeley:
Please pull the new density profile calculator, Thanks!


rainwoodman:
Looks like we still need an allreduce operation to sum up the mass per shell cross processors.

PR #32, opened 2015-07-13
Makes the last mu bin inclusive (including mu == 1.0).

rainwoodman:
Previously mu = 1.0 was always excluded; the old behavior wasn't
exactly wrong, but it breaks the compatibility with 1d spectrum(
by less than 1% in position of mean k, systematically)


rainwoodman:
Jaime is testing this change.

PR #39, opened 2015-07-21
move files into plugins/fileformats/tpm.py, as plugins

rainwoodman:


rainwoodman:
This is a WIP.

rainwoodman:
This will be revisited after #51

rainwoodman:
This is done differently. but we have readers now.

PR #50, opened 2015-08-25
Merge HDFPainter with PandasPaintTextPainter into PandasPainter.

rainwoodman:
Pandas HDF5 is not really a true HDF5, and these two painters are essentially the same.
after reading the data from disk to the DataFrame.

Does pandas read other type of input data?

Do you think if this is a reasonable change?


nickhand:
Hmm, I think they are both useful. At least for my use case, I have galaxy catalogs stored as data frames which I've saved as pandas HDF5 files, which allows for quick I/O performance. And then separately, I have catalogs (usually from other people) that are stored as plain text files, and I've found the pandas plaintext reader "read_csv" to be very good for this (it reads millions of lines in a few seconds), at least compared to the terrible numpy.loadtxt

I'd like to keep both, but I'm fine moving one to the contrib directory and I can manage it. Or maybe we want to make the PandasPlainTextPainter the default plain text reader and remove the numpy one?

rainwoodman:
The new Pandas reader can read both PandasHDF5 and text file; there is no loss of functionality. 

HDF5 + PandasPlainText = Pandas

I was thinking with the added -ftype parameter, in principle the Pandas reader can support any formats with a pandas.read_xxx function that returns a data frame follows the same routine, without adding in any new plugins.

If we shall still keep both HDF5 and PandasText, I propose we rename the old HDF5 to PandasHDF5; since 'Pandas' HDF5 is a subset of a 'plain' HDF5'. What do you say?

nickhand:
Oh, okay now I understand. Yes, I think it makes sense just to combine those and add a format parameter. I don't know if there are any other useful formats...I know there is read_pickle, but I think that just wraps the pickle.load call.

rainwoodman:
Well, seems like fwf or sql can be useful in some cases; They all return dataframes. let's keep our mind open. on -ftype.

PR #51, opened 2015-08-25
Add a read method to the painter plugin

rainwoodman:
We shall start migrating the painters to this infrastructure.
After the migration is done, we can just rename the painters to Readers
and move Painter.paint to power.py, or some seperate place (e.g. if it
shall be reused by a paint.py that paints points to a mesh)

I've left alone HDFPainter and PandasPlainTextPainter because of PR #50


rainwoodman:
This should not have brea

PR #62, opened 2015-09-12
Closepairbias

rainwoodman:
Reader for the Close Pair Bias Project.

It reads the hdf5 format halo catalogue and adds a Proximity column, which is defined
as the distance to the nearest halo that's most massive than a threshold.


rainwoodman:
This is not intended to run on Edison. Maybe it will but not intended.

PR #64, opened 2015-09-16
Mpilogger

rainwoodman:


rainwoodman:
Do you think if this effort is worth it?

Adding in a wrapper to the standard logger to avoid 'if rank == 0'?

rainwoodman:
Sometimes if rank == 0 interferes with the mind-flow, e.g.

```
a = comm.allreduce(mya)
if comm.rank == 0
    logger.info('a = %g" % a)
```

can be replaced by
```
   logger.info('a =%g' % comm.allreduce(mya), on=0)
```

nickhand:
sorry catching up after my qual, @rainwoodman....I think this could be useful. is the base code for the logger ready to go? shouldn't be much more work to find and edit the logging statements

PR #66, opened 2015-09-27
Refactor measurepower and power.py

rainwoodman:
The interface and logic in measurepower() is hugely simplified.
It does no more than project the 3D power spectrum to a prefered basis,
be it 1d, (k,mu), or (k, poles)

This commit paves the way to reuse the code in measurepower.py to measure
correlation function, in 1d, (k,mu) or (k, poles).


rainwoodman:
@nickhand : Could you take a look that this does not break your code?

I did some ludimental tests that the numerical difference in the power spectrum is less than 1e-6. But the change of the interface in measurepower() is what concerns me.

nickhand:
@rainwoodman, yes I'll take a detailed look at this. Flying back to SF today, so either tonight or tomorrow. Looks really good after just a quick glance!

PR #80, opened 2015-11-25
corr.py does multipoles now -- needs latest version of  kdcount

nickhand:
also added some tests for the corr.py calls


rainwoodman:
Some small tweaks for python 3 are needed. I am adding them in as individual commits.

PR #82, opened 2015-11-30
Duplicate transfer and painter to len(field) if a scalar is provided.

rainwoodman:
Does this make sense to you?


nickhand:
yep, I had the same fix

PR #89, opened 2015-12-02
logic changes in compute_3d_power to track real and complex

nickhand:


nbodykit/measurestats.py
```diff
@@ -460,24 +455,34 @@ def project_to_basis(comm, x3d, y3d, edges, los='z', poles=[], symmetric=True):
         xsum.flat += numpy.bincount(multi_index, weights=scratch.flat, minlength=xsum.size)
     
         # take the sum of weights
-        scratch[...] = 1.0
-        # count modes not in singular plane twice
-        if symmetric: scratch[:, nonsingular] = 2.
+        scratch[...] = 1.0 
+        if symmetric: scratch[:, nonsingular] = 2. # count modes not in singular plane twice
```
rainwoodman: Suppose we are dealing with a pure imaginary x3d, and LOS direction is z.
If we add up positive and negative mu into the same bin, the scratch shall cancel out.

nbodykit/measurestats.py
```diff
@@ -460,24 +455,34 @@ def project_to_basis(comm, x3d, y3d, edges, los='z', poles=[], symmetric=True):
...
+                    weighted_y3d.real[:, nonsingular] = 0.
+                    weighted_y3d.imag[:, nonsingular] *= 2.
+                else:
+                    weighted_y3d.real[:, nonsingular] *= 2.
+                    weighted_y3d.imag[:, nonsingular] = 0.
+                    
+            weighted_y3d *= 0.5*(2*ell + 1.)
+            ysum[iell,...].real.flat += numpy.bincount(multi_index, weights=weighted_y3d.real.flat, minlength=Nsum.size)
+            if numpy.iscomplexobj(ysum):
+                ysum[iell,...].imag.flat += numpy.bincount(multi_index, weights=weighted_y3d.imag.flat, minlength=Nsum.size)
```
rainwoodman: If poles == [], ysum is uninitialized.. This seems to break regular P(k,mu).

nbodykit/measurestats.py
nickhand: Do you mean pure imaginary y3d (not x3d) ? We square x3d so there so issue with positive / negative canceling it (I think). I don't think I see where scratch would cancel.

nbodykit/measurestats.py
nickhand: I don't think I see what you mean. ysum is initialized on line 415 to be either real, or complex, depending on the type of the input y3d. And even if poles == [], we always compute ell = 0 in order to do P(k,mu)

nickhand: I'm still on a bunch of cold medicine so I'm probably missing something :)

rainwoodman:
Let me look thought this more carefully then.

nickhand:
for what it's worth, all of my velocity correlator tests look normal so far

rainwoodman:
I see your points now. I made a few tweaks in the comments. will push after merging.

PR #90, opened 2015-12-02
fixes output bug for P(k,mu) + improvements

nickhand:
the current logic for output: 

* output complex power as successive columns
* update column names in header by appending "_real" and "_imag"
* the 2D reader looks for "_real"/"_imag" pairs in column names and reconstructs
the complex output, which it returns
* the 1D reader just does a numpy.loadtxt, so it returns an array with real/imag as successive columns


rainwoodman:
Would you rather prefer '.real' and '.imag'?

PR #92, opened 2015-12-03
removing factor of 1/2 in ell pre-factor; 1D reader returns cols as meta

nickhand:

### DIFF
--- a/nbodykit/measurestats.py
+++ b/nbodykit/measurestats.py
@@ -481,7 +481,7 @@ def project_to_basis(comm, x3d, y3d, edges, los='z', poles=[], symmetric=True):
                     weighted_y3d.real[:, nonsingular] *= 2.
                     weighted_y3d.imag[:, nonsingular] = 0.
                     
-            weighted_y3d *= 0.5*(2*ell + 1.)
+            weighted_y3d *= (2.*ell + 1.)
             ysum[iell,...].real.flat += numpy.bincount(multi_index, weights=weighted_y3d.real.flat, minlength=Nsum.size)
             if numpy.iscomplexobj(ysum):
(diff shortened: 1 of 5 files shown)


nbodykit/measurestats.py
```diff
@@ -481,7 +481,7 @@ def project_to_basis(comm, x3d, y3d, edges, los='z', poles=[], symmetric=True):
                     weighted_y3d.real[:, nonsingular] *= 2.
                     weighted_y3d.imag[:, nonsingular] = 0.
                     
-            weighted_y3d *= 0.5*(2*ell + 1.)
+            weighted_y3d *= (2.*ell + 1.)
```
nickhand: I don't think we need this factor of 1/2 since we are already averaging mu and -mu into one bin -- agree, @rainwoodman ?

rainwoodman: If you mean we are double counting the number of modes at line 460, then yes.

But the number of modes also included non-independent modes. so sqrt(Nmodes) is no longer the error. Right?

nickhand: I think the error is still sqrt(Nmodes) . We are counting all of the modes in the box, which I think is the number you want to use, since the power at mu and -mu doesn't have to be the same in the most general case (I think). 

I have verified that this change reproduces the same power as before for the test cases, although the mode counting produces a different result. I think the version now is correct: the number of modes in a k-bin of width dk agrees with the analytic estimate N = 4π k^2 ∆k V / (2π)3.

rainwoodman: I agree with your argument for now. but we shall consult with some experts on this later.

rainwoodman:
Shall we merge?

PR #95, opened 2015-12-05
Massaging names of modules.

rainwoodman:
- removed plugins directory. There is now a plugins.py that provide
ways loading plugins.

- added extensionpoints module; plugins shall register at these
extensionpoints. The extensionpoint provides a way to create an
instance of the plugin.

- move storage classes to storage directory.

consider moving selectionlanuage away from utils and kill that directory.


rainwoodman:
How do you feel about this? @nickhand

nickhand:
I think I like this -- what's the conceptual difference between plugins and extensionpoints, i.e., why is ArgumentParser imported from plugins while everything else if from extension points?

Also, it looks like this broke FOFgroups.py

rainwoodman:
This is what I thought: 

- A plugin is plugged into an extensionpoint. Multiple plugins can be plugged into the same extensionpoint.

- plugins.py contains only the code for 'managing' the physical plugins files -- which files to read, how to load and execute the module etc.

- A slight weirdness is that the actual instantiation of a plugin is still accessed from the extension point the plugins are inserted into. But thinking about this twice it actually made sense -- only the extension point knows how to select which plugin to instantiate.

nickhand:
This sounds reasonable to me!

rainwoodman:
I will merge this then.

PR #104, opened 2015-12-09
adding in Grid datasource and GridPainter

nickhand:


rainwoodman:
Hmm. How do I generate a grid file?

nickhand:
Right now it's just the flattened Nmesh x Nmesh x Nmesh `real` array saved as a binary file, it would be easy to generate a density field with no correlations for testing purposes

rainwoodman:
Yes. Alternatively, we can add a tool that produces a density grid from a data source and a painter argument?

nickhand:
Yes, this would be useful I think. It's not exactly trivial to reconstruct
the full (Nmesh,)*Ndim array from the individual ranks (assuming we want to
save the full grid as one binary file)

PR #110, opened 2015-12-11
introduces TransferFunction plugins, YAML file parsing in power.py

nickhand:
* power examples have been updated to use YAML syntax -- CI won't succeed until latest pmesh change updated but works on laptop 
* I've implemented a basic first pass at doing "help", "defaults", "choices" in the YAML parser in power.py -- maybe @rainwoodman can take a look and spice it up a bit

A few things to note: 
   * in the ``construct_fields`` function which makes the tuples of (DataSource, Painter, Transfer), I also had to pass in any `X` options and load them, so that the X options can be specified in the config file (in addition to the command line)
   * Maybe we want to make the default Transfer chain [NormalizeDC, RemoveDC, AnisotropicCIC]. It seems to be the the desired chain in nearly all of the cases...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_install:
 install:
     - conda create --yes -n test python=$TRAVIS_PYTHON_VERSION
     - source activate test
-    - conda install --yes numpy=$NUMPY_VERSION nose cython mpi4py h5py pandas scipy pytables yaml
+    - conda install --yes numpy=$NUMPY_VERSION nose cython mpi4py h5py pandas scipy pytables pyyaml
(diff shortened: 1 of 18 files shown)


rainwoodman:
So, this morning I am just thinking what if we use YAML just for the inputs, but still use argparse for modes etc?

nickhand:
Sorry, you mean use YAML for just the input fields, and use argparse to parse the rest of the parameters?
That would require possibly two parameter files, right?

rainwoodman:
Indeed. That's a bad idea. I would prefer keeping the commandline in parallel with the parameter files -- still feel to write a file a measurment is kind of a mental burden. 

Is this ready to be merged? I can look at extending the commandline / merge with yaml parser later (after the fastPM paper is done.)

PR #113, opened 2015-12-16
introducing DataSet instead of PkmuResult; renamed plaintext readers

nickhand:
* see the documentation in dataset.py for examples
* useful subclasses of `dataset.DataSet` are `Corr1dDataSet`, `Corr2dDataSet`, `Power1dDataSet`, and `Power2dDataSet`
* renamed `files.ReadPower1DPlaintext` and  `files.ReadPower2DPlaintex`t to just `Read1DPlaintext` and `Read2DPlaintext`


nbodykit/dataset.py
```diff
@@ -0,0 +1,849 @@
...
+        
+    @property
+    def variables(self):
+        """
+        Alias to return the names of the variables stored in `data`
+        """
+        return list(self.data.dtype.names)
+        
+    @classmethod
+    def __construct_direct__(cls, data, mask, **kwargs):
```
rainwoodman: Could you also write in the docstring the difference from object initalized via `__init__` ?

nbodykit/dataset.py
```diff
@@ -0,0 +1,849 @@
...
+        
+        data = self.data.copy()
+        mask = self.mask.copy()
+        for i, idx in enumerate(indices):
+            data = numpy.take(data, idx, axis=i)
+            mask = numpy.take(mask, idx, axis=i)
+        
+        return self.__finalize__(data, mask, indices)
+        
+    def to_pickle(self, filename):
```
rainwoodman: Why not the standard pickle protocal?

nickhand: I'm just pickling the components needed to reconstruct the class -- mainly to avoid issues with pickling while the code is still in flux, i.e., the class name changes. I've been burned by that before, but I'm open to just doing the standard pickling

nbodykit/dataset.py
```diff
@@ -0,0 +1,849 @@
...
+         a dict holding the bin edges for each dimension in `dims`
+        
+    coords : dict
+        a dict holding the bin centers for each dimension in `dims`
+    
+    force_index_match : bool
+        if `True`, when indexing using coordinate values, return
+        results for the nearest bin to the value specified
+        
+    sum_only : list of str
```
rainwoodman: Is there a more descriptive name than sum_only? It is a list of fields do not need to be averaged. At least `sum_only_fields` would be less suggestive of an action.

A possible extension to this list is a dictionary, fieldname: attribute. and attribute can be SUM_ONLY. It requires more typing though.

nbodykit/dataset.py
```diff
@@ -0,0 +1,849 @@
...
+    dims : list
+        a list of strings specifying the names of each axis of the input data
+    
+    edges : dict
+         a dict holding the bin edges for each dimension in `dims`
+        
+    coords : dict
+        a dict holding the bin centers for each dimension in `dims`
+    
+    force_index_match : bool
```
rainwoodman: fuzzy_index?

nbodykit/dataset.py
nickhand: do you mean for `__construct_direct__`? it's designed for internal use only, it's needed for when you slice the DataSet, in order to build the new, sliced class easily

rainwoodman: Yes. I know it is for internal use only... but just by looking at the code I can only tell when it used, but cannot easily tell the difference between the object it returns with an object created by DataSet(....). I worry after a month you'll get confused too..

nickhand: ah, I see. I'll add some more documentation. The object returned by the class should be indistinguishable from the object returned by DataSet(...) (I think)

nbodykit/dataset.py
rainwoodman: Will this do? This way only the necessary information is saved. -- not sure if I understand the issue correctly.

```
def __reduce__(self):
    nesseary = .....
    d = { ....}
    return (Dataset.__unpickle__, (d, ))
@classmethod
def __unpickle__(d):
    ....
```

rainwoodman: Oops I see now what you mean the class name changes... I'd still prefer to have a standard pickle for things like sending a Dataset via MPI.

nickhand: does the "__unpickle__" function need to be defined at the module level in order to be pickled? That's the only way I can get the above to work without a "cannot pickle instance-method" error

PR #125, opened 2016-02-01
first pass at implementing `Algorithm` extension point

nickhand:
Only working example so far replaces corr.py. Just want to see what @rainwoodman thinks before doing the major code overhaul. I think this is a definite improvement over what we have

the only new commit is 0046a1b. see message there for more description.

### DIFF
--- a/nbodykit/extensionpoints.py
+++ b/nbodykit/extensionpoints.py
@@ -14,7 +14,8 @@
     
 """
 class PluginInterface(object):
-    """ The basic interface of a plugin 
+    """ 
+    The basic interface of a plugin 
     """
     def initialize(self, args):
         ns = self.parser.parse_args(args)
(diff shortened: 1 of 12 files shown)


nbodykit/extensionpoints.py
```diff
@@ -336,8 +337,180 @@ def open(self):
...
+        args, kwargs = klass._namespace_to_args(ns)
+        return klass(*args, **kwargs)
+       
+                    
+def AlgorithmExtensionPoint(cls):
+    """ Declares a class as an extension point """
+    return add_metaclass(AlgorithmPluginMount)(cls)
+    
+@AlgorithmExtensionPoint
+class Algorithm:
```
rainwoodman: I've been thinking of a 'prepare' or 'initialize' stage for 'DataSource' (to read in BoxSize, set up units etc). Would something like this be useful for an algorithm as well?

nickhand: Yeah I think it would. One issue we'll need to think about more with the Algorithm is if we want to do anything fancy to handle differences in types between what goes on the command-line and is parsed by the parser and the arguments that would be passed directly to the Algorithm's __init__. Something like a prepare or initialize would help in this case I think

rainwoodman: I wonder if it is possible to 'repurpose' Python's standard __init__ for
this, since the 'default' initialization can be placed with __new__.

PR #127, opened 2016-02-03
Adding PeriodicPower with working YAML reader and cmd-line parser combo

nickhand:
This PR implements:

* `ReadConfigFile` action in `nbodykit.plugins` that reads a YAML file and checks various things using the underlying parser actions
* plugins.ArgumentParser returns namespace from YAML file when passed via `-c`, `--config`; the preparser also loads any plugins specified with `X` in YAML file
* PeriodicPower algorithm now working
* power examples updated to use nbkit.py PeriodicPower and the -c syntax (or @filename for one case) 
* no changes so far to power.py -- everything there should still work

### DIFF
--- a/bin/nbkit.py
+++ b/bin/nbkit.py
@@ -35,9 +35,13 @@ def main():
     subparsers = parser.add_subparsers(dest='algorithm_name')
     for s in valid_algorithms:
         
-        # copy the parser for each Algorithm
+        # make a subparser for this sub-command
         subparser = subparsers.add_parser(s)
-        subparser.__dict__ = getattr(algorithms, s).parser.__dict__
+        
+        # make the subparser the same as the algorithm parser
+        p = getattr(algorithms, s).parser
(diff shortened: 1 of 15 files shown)


rainwoodman:
I did some refactoring work -- merging AlgorithmPluginMount with PluginMount. 

Would it be easier if you rebase this on #128, since that PR reduced the size of the code base?

nickhand:
There are a couple of small fixes for this PR I was working on so let me pull your changes into my local branch and I'll put in a PR for everything

rainwoodman:
A few more points. I realized argparse also describes the type of the kwarg
so it makes more sense to go from argparse to defaults of --init-- than
splitting them to two places.

Then plugins.argument parser really is "auto load argument parser".  It
shall only be used by nbkit.

Yu

nickhand:
Okay, I think I see what you're trying to do. Let me rework the config parsing a bit to match this new API

PR #129, opened 2016-02-03
Merging #128 and fixing config parser

nickhand:
* this closes #128 and fixes the config parser to reflect the changes in `nbkit.py`
* the config file reading is handled as a custom action where the parser will ignore all errors until the end, at which point it will return the config namespace, or crash if the config file isn't present
* the argument parsers for each plugin have to be from `plugins.py` to handle this custom config YAML parsing
* I added a `preparse_from_config` keyword to the plugin ArgumentParser which will read these elements from the config file and return them as part of the namespace . This is important for the `output` option in `nbkit.py` -- we can set `required=True` if we want to specify it in the config files, so check for it manually

### DIFF
--- a/bin/corr.py
+++ b/bin/corr.py
@@ -47,7 +47,7 @@ def initialize_parser(**kwargs):
 
     # add the input field types
     h = "one or two input fields, specified as:\n\n"
-    parser.add_argument("inputs", nargs="+", type=DataSource.create, 
+    parser.add_argument("inputs", nargs="+", type=DataSource.fromstring, 
                         help=h+DataSource.format_help())
 
     # add the optional arguments
--- a/bin/fof.py
+++ b/bin/fof.py
@@ -22,7 +22,7 @@ parser = ArgumentParser(None,
         )
 
 h = "Data source to read particle position:\n\n"
-parser.add_argument("datasource", type=DataSource.create,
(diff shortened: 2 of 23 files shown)


nickhand:
Everything seems to be working now so I merged this.

The handling of the `-c` YAML reader is probably not optimal, but it is working now and I think it's a good first start so far.

rainwoodman:
I am a bit wary that extensionpoint.py also pulls in plugins.ArgumentParser for the plugins. Would that mean '-c' per plugin?

rainwoodman:
It also looks strange that ArgumentParser takes args as an constructor argument.

rainwoodman:
A slight refactor of TaskManager can get rid of args in constructor. 
The args shall not be parsed before parse_args / parse_known_args is called. 
Let me try to clean this up today.

nickhand:
I agree I don't think it's optimal at the moment. In the current setup, the user has to add "-c", "--config" to the plugin parser explicitly. The issue I was running into is that the PeriodicPower algorithm needs to tell the ConfigParser how to read the `fields` from file (which is done by the plugin parser), so the plugin parser needs to do the '-c' option. 

Will you be at the lab today? We can chat about this some more in person

rainwoodman:
I will head to the lab for the lunch talks soon. 

What about insisting ConfigParser knows how to read the fields?

PR #134, opened 2016-02-05
comm is set for plugins through global variable

nickhand:


nickhand:
@rainwoodman, there looks like there a few bigfile related issues with the tests...will you take a look?

rainwoodman:
I am on this.

nickhand:
thanks! looks like all the tests have passed now

PR #137, opened 2016-02-12
Update README.rst

melissajoseph:


rainwoodman:
This $PWD is hard to be missed than '.'.

PR #140, opened 2016-02-15
Implementing the Bianchi multipoles estimator for (ra, dec, z) surveys

nickhand:
this is a pretty PR -- it adds a `BianchiPower` algorithm which can take (ra, dec, z) for a "data" and "randoms" catalog and measure the ell=0, 2, 4 multipoles using FFTs

the new codes:
* I've created a `Cosmology` extension point, which one plugin so far, based on `astropy.Cosmology` (see da3702b)
* BianchiPower: an algorithm to use FFTs to compute power spectrum multipoles from (ra,dec,z) data with survey geometry
* RaDecRedshift: very bare DataSource (with a bad name) that uses Pandas `read_csv` to handle reading of (ra,dec,z) (possibly with a `bunchsize`, since the randoms file can be large) 
require a BoxSize
* TracerCatalog: the main DataSource needed by BianchiPower, which includes the "data" catalog as well as the "randoms" catalog; it sets up the cartesian box, handles FKP weights, and computes n(z)
* FKPPainter: given a data and randoms, paint the weighted density field from Feldman, Kaiser, Peacock: N_data - alpha*N_randoms


nickhand:
thoughts on a better name than BianchiPower, @rainwoodman? Maybe SurveyPower or WindowedPower?

rainwoodman:
- Bianchi is the first author? What about adding the publication year too? 

- Are we solidly depending on astropy? The 'sample' means we can use something really slow to tabulate the function once then use interpolation... which means we can probably do something pure python, e.g. https://github.com/rainwoodman/lyamock/blob/master/lib/cosmology.py#L60

nickhand:
* hmm yes, we could go with Bianchi2015Power. I was a little worried that it would be unclear as to the real use (for surveys with selection functions), but I guess something like SurveyPower really isn't any clearer. There's probably not a simple way to communicate everything in just the name

* at the moment, I've set it up such that `astropy` is only required if you choose  to use `AstropyCosmology` (similar to `Pandas` readers). We could add in a default python-only cosmology plugin for simplicity. I do have such a function lying around (but generalizing it to say, w != -1 or massive neutrinos is the issue). 

My thinking on astropy was: 
* I think the `Cosmology` class is relatively stable now
* it comes included in the default Anaconda distribution
* it can handle massive neutrinos and w = -1 
* if we want to support functions other than `comoving_distance`, those will already be implemented in astropy, so we just wrap it

rainwoodman:
- What about BianchiFFTPower?

- If the cosmology class in astropy is relatively stable, then shall we consider not using a plugin but directly use the cosmology object in astropy? I am also a bit wary about going too far on the 'everything is a plugin' road.

nickhand:
* okay I like that name better
* I had similar thoughts about the "everything is a plugin" aspect. I think
we could just make it part of the base code, although I think we would
still want to keep a lot of the same syntax of the PluginInterface, i.e,
the colon separated strings to use on the command-line. I guess we could
just move the existing code to the base code, or just pull the features we
want out of PluginInterface

rainwoodman:
It sounds like we would like to have something that allow us to create an object from the command line easily -- let's stick with a plugin before a better cmdline parser is written?

nickhand:
yes, I think what we need is a unified parser to handle the command-line interface of classes + the configuration file reading, and then that command-line interface can be applied to plugins and non-plugins. I agree we can stick with the cosmology plugin for the moment...

PR #143, opened 2016-02-18
adding a possible Schema + init decorator solution

nickhand:
I think this solves a bunch of issues at once

* see `cosmology.Cosmology` as an example of how this gets implemented
* the schema (a list of named tuples) gets attached to the __init__
* schema is updated with calls to the @Argument decorator
* do not need to pass `default` or `required` values to the @Argument decorator, as
it can be inferred from the __init__ signature
* the @Configure call auto-initializes the class instance with the arguments passed
to the __init__, using the schema to cast
* then, the body of the __init__ becomes the same as `finalize_attributes()`
* docstring is updated by the @Configure call (which is last, and the schema is full)

We can set up the plugins so the metaclass calls `plugin.register()` (or maybe `plugin.configure()` now?) and then wraps the __init__ with the Configure decorator so the same syntax can apply to the plugins automatically

There are obviously tweaks to be made, but I think the decorator solution is pretty powerful


nbodykit/extensionpoints.py
```diff
@@ -145,78 +88,98 @@ def __init__(cls, name, bases, attrs):
...
-                                        add_help=False, 
-                                        formatter_class=cls.help_formatter)
-
-            cls.parser._print_message = _parser_print_message
+        if cls.plugin_name not in cls.registry:
+
+            # initialize the schema and alias it
+            if cls.__init__ == object.__init__:
+                raise ValueError("please define an __init__ method for '%s'" %cls.__name__)
+            cls.__init__.__func__.schema = ConstructorSchema()
```
rainwoodman: Seems like we can remove this ```__func__``` ? It helps python 3 too.

nickhand: Yes, I'll be adding an update to this PR implementing  sub fields in the
schema as well as a fix for this

nickhand:
okay, looks like we are back to passing on all tests. shall I merge this?

PR #153, opened 2016-03-01
initial commit of HOD class which tests out halo tools

nickhand:
the idea is that this new DataSource can:
* load a FOFGroups halo catalog
* use halo tools to populate the mock
* return the central/satellite galaxy position / velocity


nbodykit/plugins/datasource/Zheng07HOD.py
```diff
@@ -0,0 +1,177 @@
...
+        
+        # build the model and set defaults
+        base_model = em_models.PrebuiltHodModelFactory('zheng07')
+        self.model = em_models.HodModelFactory(baseline_model_instance=base_model, 
+                                                satellites_profile=sats_prof_model)
+        for param in self.model.param_dict:
+            self.model.param_dict[param] = getattr(self, param)
+        
+        # data from the halo catalog
+        [[Position, Velocity, Mass]] = self.halocat.read(['Position','Velocity', 'Mass'], full=True)
```
rainwoodman: I am bit worried about parallelism. If halotools requires global counting (e.g. to normalize the total number of halos) we may need to gather the data to a single rank and run from there. We may want to check with Andrew about this.. @aphearin

nickhand: That is a good point. My brief initial look at the algorithms seemed to indicate we could populate chunks of halos in parallel (the total number of galaxies is stochastic anyways).

aphearin: Hey, I think I might need a little more background to understand the problem. What I can say is that I have not parallelized the mock population, basically because runtimes are so fast there has been no need. For example, for the Bolshoi simulation, it only takes ~250ms on my laptop to populate a mock with Zheng07 (not counting the very first time, which is slow). This part of the calculation scales roughly linearly with Nhalos, so for example Multidark takes ~8x longer to populate as there is 8 times more volume.

rainwoodman: That's exactly the point -- nbodykit is parallel, but halotools is not, so we need to adapt between the two. What we have here is each rank calls halotools with its local partition of the halos, and obtain a local set of galaxies.  (scenario 1, current implementation)

The alternative is all halos are gathered to a single rank, then halotools is called, then, finally the produced galaxy catalogue is distributed to the ranks. (scenario 2, serialized implementation)

Are these two processes equivalent? I worry about correlation of random number generators, and the possibility of missing global constraints with scenario 1. I agree runtime is not a concern, but correctness is.

aphearin: Ok, that's clarifying. With Scenario 1, there are several things that you need to be careful about. For example, how is Lbox being set for the subvolumes? Is it the size of the partition, or is it the periodic boundary conditions of the full snapshot? This is important because if the `enforce_periodic_boundary_conditions` function detects points (satellites) beyond Lbox, their positions are wrapped back inside box. Scenario 2 does not have this problem and is generally simpler as far as Halotools is concerned, but I understand why you'd rather not suffer the runtime hit of recollecting and rescattering (or is this not a worry?)

As for correlation of random number generators, I have not had to worry about this before now so this is untested. I'd be discuss how to experiment with this for your use case though.

nickhand:
@aphearin -- I am wondering if we are setting the sim defaults using the API you envisioned,  [see here](https://github.com/nickhand/nbodykit/blob/7f6a5b78764ac69c5b956e38eb09efa992a6255b/nbodykit/plugins/datasource/Zheng07HOD.py#L46-L48). In particular, it seems like we could pass redshift and cosmology to individual model components, but is there way to pass `Num_ptcl_requirement` to the MockFactory, when calling `populate_mock` from the model instance?

aphearin:
@nickhand -- so far as mock population with a Zheng07 HOD is concerned, the cosmology never comes into play, so I think whatever you do with this should be fine. As for `Num_ptcl_requirement`, is the issue that you are interested in the impact for various choices of the minimum halo mass, or is it that you just want to change the default behavior of the MockFactory?

nickhand:
@aphearin -- at the moment, we aren't concerned with minimum halo masses, so I wanted to include all the halos by default. I think this can be done with the `apply_completeness_cut` keyword of the HodMockFactory, but I wasn't sure if there was an existing way to pass that keyword down to the automatically-created HodMockFactory when calling `populate_mock` from the model instance

aphearin:
@nickhand - your code was well-organized and it showed me how you were having to bend over backwards to adjust the particle completeness cut, so I made a few API changes:

https://github.com/astropy/halotools/pull/410

This gave me an excuse to clean up some code I'd been meaning to streamline for a while (particularly `populate_mock`, which had some nasty nested control flow that was hard to follow). Have a look at this PR and let me know what you think, I think this should cover your use-case a lot more cleanly, but do let me know if you see room for improvement.

PR #163, opened 2016-04-22
Add automated documentation

rainwoodman:
The current problem is the document is very sparse.

See http://rainwoodman.github.io/nbodykit/


nickhand:
this looks like a good start!

do you know why some algorithms aren't showing up (BianchiFFTPower, FiberCollisions,)? was it built on an old version of the code?

I need to familiarize myself with the sphinx documentation syntax...seems like some of the formatting of the current docstrings is messed up. Then I'll start adding documentation...

And do we want to use readthedocs rather than github pages? seems like that is the more popular choice these days (although this is another area I need to familiarize myself with)

rainwoodman:
Bianchi is added -- I forgot rerunning apidoc.sh after the rebase.

I believe numpy styled doc string is supported (I slightly favor numpy style than the alternatives, for no good reason): https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

I haven't used readthedocs before. We need to decide on a theme too. The 'nature' scheme has poor highlighting rules...

PR #171, opened 2016-05-16
RTD back to building successfully; have to mock a bunch of modules

nickhand:
* the ``import nbodykit`` means the custom modules have to be there
* get around this using the mock modules


rainwoodman:
It's miracle this actually works.. -- lucky that we never call any dependdency in module level in nbodykit.

PR #180, opened 2016-06-03
potential build framework with nightly cron jobs

nickhand:
#### Build framework
* the setup right now builds a `stable` source tar which tracks the `master` HEAD and a `latest` tarball which tracks the `develop` branch
* in `/usr/common/contrib/bccp/nbodykit`, there is a `nbodykit-latest.tar.gz` and `nbodykit-stable.tar.gz`, as well as `activate.sh`
* the intended usage is then `source /usr/common/contrib/bccp/nbodykit/activate.sh`. If no arguments are provided, the `stable` tar ball is loaded and `latest` and `stable` can be passed to the script too. 

#### Nightly builds

* the script `nightly-build.sh` installs the head of develop and master from github and installs the dependencies as well -- the dependencies take ~10-20 minutes to install because of `pfft-python` so eventually it would be good to implement some sort of version checking
* `bash nightly-build.cron` will add a cron job that runs `nightly-build.sh` at midnight each night. you can run this on edison and cori independently to add to the crontab for each machine 
* I've set the crontab up so we will see how well it works -- if we want to get really fancy, we can send emails if the script fails for some reason

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 examples/data
 examples/nbodykit-data.tar.gz
 examples/output
+nersc/cori
+nersc/edison
--- a/nersc/activate.sh
+++ b/nersc/activate.sh
@@ -1,5 +1,21 @@
 #!/bin/bash
 
+# can supply 0 or 1 argument
+if [ "$#" -gt 1 ]; then
+    echo "usage: activate.sh [latest|stable]"
+    exit 1
+fi
+
+# if no version provided, use 'stable'
+if [ $# -eq 0 ]; then
+    version="stable"
+elif [[ "$1" -ne "stable" || "$1" -ne "latest" ]]; then
+    echo "valid version names are 'stable' and 'latest'"
+    exit 1
+else
+    version=$1
+fi
+
 if [[ -n $BASH_VERSION ]]; then
     _SCRIPT_LOCATION=${BASH_SOURCE[0]}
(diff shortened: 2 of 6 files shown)


rainwoodman:
Looks good to me. I am slighly concerned about relative path in the script name (_SCRIPT_LOCATION). I recall it wronged me once, hence I had a readlink there. 

Bash has a builtin to obtain the abspath I believe.

Also, I remember NERSC will email you automatically about the cron job.

nickhand:
you're referring to this in `activate.sh` ? 

if [[ -n $BASH_VERSION ]]; then
    _SCRIPT_LOCATION=${BASH_SOURCE[0]}
elif [[ -n $ZSH_VERSION ]]; then
    _SCRIPT_LOCATION=${funcstack[1]}
else
    echo "Only bash and zsh are supported"
    return 1
fi

did you have a readlink there before? I don't think I edited that part

nickhand:
looks like everything worked last night with the builds. 

one small issue. Cron jobs are specific to the login node, which took me a while to realize, so I have multiple cron jobs submitted on separate login nodes. Do I need to contact support to clear them, or is there another way, outside of randomly getting the right login node?

rainwoodman:
Probably contacting the support. I thought there was only one shared cron per machine.

rainwoodman:
It's later calculating DIRNAME.

PR #182, opened 2016-06-07
testing overhaul using py.test and pytest-pipeline

nickhand:
* converted all of the existing bash tests to py.test tests using pytest-pipeline to run the pipeline
* caching data in ``$HOME/.nbodykit``, as downloaded from nbodykit-data directory
* ``nbodykit-data`` also stores the correct output files and the tests verify the correctness of the new output files


nickhand:
So it seems like doing comparisons of the output files using MD5 sums won't work for the power/corr files due to numerical issues -- some of the power values are only equal to one decimal place when comparing the Travis results to my laptop. This only seems to effect the 'power' columns of the output files, so likely a numerical issue related to the FFTs. 

@rainwoodman, thoughts on this? It's probably best to do the comparison by actually reading in the files into DataSet objects and checking to the desired accuracy...

rainwoodman:
What about reading in the file with loadtxt then do a 'allclose'? 
How stable is the PowerxdDataset API?

nickhand:
Yes I think we'll have to do something like that, although from my initial inspection some of the results actually have fairly large discrepancies, which seems odd to me. Do you have a feel for where these differences come from on different machines?

rainwoodman:
I don't.. Are you confident the version of code is the same? Did you run with the same number of cores?

nickhand:
okay I've got this sorted! I figured out that all of the imaginary columns are essentially noise, since they are consistent with zero, so I'm ignoring those columns in the relevant tests

PR #186, opened 2016-06-09
adds proper versioning for release 0.1.1

nickhand:


nickhand:
@rainwoodman -- what do you think of this logic? 

It's borrowed from xarray and halotools/astropy seem to have similar logic. It uses setup.py to create a `version.py` in the main directory. 

Basically, I think we should set `ISRELEASED` to True on the master branch to use the X.X.X versioning and then on the develop branch, if `ISRELEASED` is False, it will append the `-dev-xxxx` where xxxx is the latest git hash. This setup behaves really well with readthedocs so it seems to make sense.

rainwoodman:
I wonder why we need to carry an accurate version number on the development
branch -- and given that if setup.py is not ran the number is out of date,
I sort of feel it will be very frequently out of date.

I think we need this because of the develop/master split.

In order to justify having a develop branch, we require master to be
'stable'.
Because we require master to be stable, then release and version umbers are
no longer useful, because every commit on master is a stable release.
Then we need to version frequently -- so frequent that we shall add hash
ids to the version, which in turn inflates the build infrastructure,
complicates the merge, etc etc.

xarray and halotools are very young packages. Besides, by physists. I am
not sure if we really want to follow them as bibles.

In fact, I don't see what's wrong with a single, linearly evolving master
branch that is frequently tagged / released into maintenance branches.

Real projects -- pip, gtk, scipy, numpy, jupyter in fact, every major
packages out there all follow this model.

1. https://github.com/GNOME/gtk/ This thing has been backing the  linux
desktop business for 20+ years, and still very actively developed. They do
not have automatic git hash tagging -- actually, they do not seem to have a
'develop' branch.

2. https://github.com/pypa/pip has only one branch. They handle version
with __init__.py -- which is more deterministic than with a generated
version.py, see

https://github.com/pypa/pip/blob/master/setup.py#L34

What about we follow the pip example, use a manually maintained version
number in __init__.py, which goes like 0.1.0 and increases very frequently.
The document generation tool may need to be tweaked a bit for 'current'
documentation. But if we only have one master branch, then it should be
quite easy to deal with.

rainwoodman:
pip may have recently just got rid of there 'develop/master' branch model...

See

https://github.com/pypa/pip/commit/b095e41a2a7407048d7a7892c885b9dc23cf5058

nickhand:
Okay, I agree with your points here. Better to simplify things when we can! 

So we'll merge develop back into master, tag releases with the X.X.X convention, and then continue development by bumping the version and adding back '.dev0' 

what's the meaning '0' in .dev ? we could in theory increment that number for development releases?

nickhand:
We will also need to manually set the latest `stable` version when doing the nightly builds on NERSC, so that we can pip install using the right tag. Shouldn't be too hard to do. 

Perhaps we should make a development doc so we don't forget all of the steps.

rainwoodman:
pip complains if there is no 0. The idea is the number can be increased for `development releases`.

I'll merge #190 and write a maintainer.rst describing the process.

PR #190, opened 2016-06-10
Simple version

rainwoodman:


nickhand:
this looks good to me! 👍

rainwoodman:
I added this file to the documentation

https://github.com/bccp/nbodykit/blob/master/docs/maintainer.rst

We'll do stable, unstable, and a nightly build at nersc. I'd like this to be automatically triggered with a github hook or as part of the travis build script (if it is successful, notify nersc to rebuild)

Do you want me to look into this? We need to store a encrypted nersc credential on Travis or somewhere.

nickhand:
This maintainers file looks good. If I'm reading it right, you want to have
master, stable, and unstable branches? And what do you mean by reset the
branch to a specific tag?

Triggered builds on nersc would be great. Go ahead and take a look at that.

Also, FYI I've finally got a bit more documentation together that I'll be
adding soon. I've got the doc build warnings down from 1000s to 2!

rainwoodman:
10000 to 2. That's quite an achievement!

We can reset the branch with git reset.

The idea is tags, once created are never modified, by convention. (If I
remember correctly I read about this from a -dev mail list of a major
project.)

But branches can be moved. so we use stable/unstable branch as a movable
tag. Which also has the benefit of triggering travis automatically.

rainwoodman:
Sorry I mean `git checkout -B`

nickhand:
Okay that makes more sense. So master tracks the entire history but the
stable and unstable branches only know about the latest respective version
(if the git checkout -B erases the history). Is that right?

Maybe we should add the relevant git commands to this maintainers doc

rainwoodman:
Sure.

The auto rebuild will be a security risk regardless what scenario I can
come up with.

It essentially allows anybody with access to the repository to run any code
under your name on the nersc computers.

For now it is not an issue since we know everyone with a write permission
to the repo. But in a longer term, you probably want to disable the crontab
as well.

If we do that then we shall build stable/unstable at nersc explicitly as
part of the maintenance cycle.

rainwoodman:
If we trigger the rebuild manually, then there is not even need for stable
/ unstable branches.

We can just make a tar.gz release (dev or not), upload to pip, then write a
shell script that asks nersc to rebuild.

pip knows how to avoid dev packages.

That's even less book keeping. Did I miss something here?

- Yu

nickhand:
I think the tool we are missing is something to figure out the packages in
the tar bundles and then do the relevant pip install command if it is out
of date.

Then I think we keep can the nightly build of the master HEAD, check if we
need to updates any stable/unstable releases, and also check if we need to
update the dependencies.

nickhand:
@rainwoodman okay, so I think what I described above is pretty easy to do in python actually. If all of the packages are on pypi, we can parse the json url provided for each package, get the upload time of the latest stable release and compare it to the modified time of the latest tarball on nersc. 

Does this sound reasonable to you? We just need to decide what versions we want to build. It makes sense to use this pypi checking for the dependencies, and then we can have a nightly build of the HEAD of master. And then maybe just check for the latest stable release of nbodykit on pypi and also build a stable version. In the current framework, I am imagining the "latest" tarball to map to the HEAD of master and the "stable" version to map to the latest release on pypi. 

If this sounds okay, I can code up the new build script

rainwoodman:
1. I would think a version number is more reliable than a time stamp. Since
PYPI demands an incremental version number and doesn't allow updates.

2. I also wonder what happens if we simply unzip the bundle, set
PYTHONPATH, and do pip install -U -- will pip automatically 'do nothing' if
there are no updates?

nickhand:
from some initial tests, I think we can unpack the tarball and then do something like:

PYTHONPATH=lib/python2.7/site-packages/ pip install -U --no-deps nbodykit -t lib/python2.7/site-packages/

which should only look in the one directory for packages, and install into that same directory. We can then tar that directory and reset PYTHONPATH and remove the temporary untarring directory

rainwoodman:
Sure. If it works this is simpler than parsing json ourselves..

- Yu

PR #203, opened 2016-07-01
indexing + basic functions for selection language

nickhand:
implements a few new features in selection language: 

* vector indexing via "Position[:,0]"
*  log, log10, exp functions can be used on either side of comparison, i.e., "log10(Mass) < 12" or "LogMass < log10(10**12)"
* better exception handling -- an error is thrown if selection array is multidimensional, i.e., "Position < 100" or "Position[:,:2] < 100" will throw an error
* functionality for "is not", i.e., "Mass is not nan"
* adding a unit test suite  for selection language

This also cleans up the selection query in Pandas and Plaintext data sources (and implements new examples, so reload the cache) such that the builtin columns (Position, Velocity, etc) can also be selected

### DIFF
--- a/nbodykit/ndarray.py
+++ b/nbodykit/ndarray.py
@@ -32,7 +32,7 @@ def extend_dtype(data, extra_dtypes):
     if any(name in existing_names for name in new_names):
         raise ValueError("adding a new data type with name already present in structured array")
         
-    exsiting_dtype += extra_dtypes
+    existing_dtype += extra_dtypes
 
     # add old variables
     new = numpy.zeros(data.shape, dtype=dtype)
--- a/nbodykit/ndarray.py
+++ b/nbodykit/ndarray.py
@@ -6,10 +6,9 @@ These are really handy things if numpy had them.
 """
 import numpy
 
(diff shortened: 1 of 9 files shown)


nbodykit/ndarray.py
```diff
@@ -6,6 +6,42 @@
...
+        
+    dtype += extra_dtypes
+
+    # add old variables
+    new = numpy.zeros(data.shape, dtype=dtype)
+    for col in names:
+        new[col] = data[col].copy()
+        
+    return new
+
```
nickhand: @rainwoodman is there a better way than this to copy a structured array and append new fields to the data type?

nickhand:
@rainwoodman let me know what you think of this -- you can take a look at examples in the new unit test suite for the selection language that I've added

nickhand:
@rainwoodman shall I merge this?

nbodykit/ndarray.py
rainwoodman: I would call this grow dtype or extebd dtype. You can skip the extra copy. Iuduallt do

A[vol][:] = b[col]

Which describes what us going on.

PR #205, opened 2016-07-06
adds external module "extern" (holding six and pyparsing right now)

nickhand:


nickhand:
@rainwoodman coverage up to 84%!

PR #207, opened 2016-07-09
cleaning up memory usage in BianchiFFTPower

### LINKED ISSUE #206: BianchiFFTPower crashing on Edison
I am suddenly seeing crashes on Edison running BianchiFFTPower. The error message is mysterious:

run: error: nid00018: tasks 0-23: Killed
srun: Terminating job step 1020397.4
srun: Force Terminated job step 1020397.4

It occurs when doing the Bianchi transfer function when (i, j, k) are nonzero:

https://github.com/bccp/nbodykit/blob/master/nbodykit/measurestats.py#L133

I think this is possibly memory-related. Thoughts, @rainwoodman ?

nickhand:
* this cleans up the memory usage in compute_bianchi_poles() by removing a few extra copies 
* probably not optimal memory usage, but fixes #206 related to memory issue on Edison

### DIFF
--- a/nbodykit/measurestats.py
+++ b/nbodykit/measurestats.py
@@ -239,8 +239,6 @@ def compute_bianchi_poles(comm, max_ell, catalog, Nmesh,
     result = []
     result.append(pm.complex*volume)
     
-    print asizeof.asized(density, detail=1).format()
-    
     # the x grid points (at point centers)
     cell_size = pm.BoxSize / pm.Nmesh
     xgrid = [(ri+0.5)*cell_size[i] + offset[i] for i, ri in enumerate(pm.r)]
--- a/nbodykit/measurestats.py
+++ b/nbodykit/measurestats.py
@@ -233,11 +233,12 @@ def compute_bianchi_poles(comm, max_ell, catalog, Nmesh,
     transfer(pm, pm.complex)
     if rank == 0: logger.info('ell = 0 done; 1 r2c completed')
     
+    # save volume for normalization purposes
     volume = pm.BoxSize.prod()
     
     # store the A0, A2, A4 arrays
     result = []
-    result.append(pm.complex*volume)
+    result.append(pm.complex*volume) # this will make a copy
     
     # the x grid points (at point centers)
     cell_size = pm.BoxSize / pm.Nmesh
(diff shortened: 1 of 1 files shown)


nickhand:
@rainwoodman, do you want to take a look at compute_bianchi_poles() in measurestats to see if you see any obvious memory optimizations? This PR removes a few extra copies

rainwoodman:
Yes. I will.

nickhand:
Great, thanks! One thing I just noticed is that I am using 'f8' in the ParticleMesh for BianchiFFTPower, which could easily be set to 'f4' to halve the memory usage. I think I may have changed this from 'f4' to 'f8' while setting up the unit tests and testing the round off errors in the imaginary components of the power spectra, but I don't think it's necessary.

rainwoodman:
In my experience 'f8' is unncessary even in high res simulations like FastPM. So I'd say use f4.

nbodykit/measurestats.py
```diff
@@ -220,26 +223,26 @@ def bianchi_transfer(data, x, i, j, k=None, offset=[0., 0., 0.]):
...
-    A0 = pm.complex.copy()
     if rank == 0: logger.info('ell = 0 done; 1 r2c completed')
     
+    # save volume for normalization purposes
     volume = pm.BoxSize.prod()
     
     # store the A0, A2, A4 arrays
     result = []
-    result.append(A0*volume)
+    result.append(pm.complex*volume) # this will make a copy
```
rainwoodman: I would think 

A0 = pm.complex * volume

then result.append(A0) is more intuitive. And uses the same amount of memory.

rainwoodman:
Looks pretty frugal to my eyes other wise.

Is it possible to rewrite looping over slabs as a generic function? I remember measurepower has a similar case. Eliminating the loop from the main logic can make the code more readable, since looping is an implementation detail.

nickhand:
Do you mean something like a function `apply_slab_by_slab` that would handle the looping and hide it from the main function?

rainwoodman:
Yes. But it could be non-trivial to write a general function -- due to RSD.

nickhand:
Yeah I think it will be hard to hide the for loop generically. What we can probably do is use a bit more object-orientated code to improve the readability. I'm thinking something like a grid iterator that can do `for slab in grid` and `slab` will have functions/attributes to return k, mu, etc with the right dimensions

In the least, this will help clean-up messiness related to broadcasting and indexing of the `x` and `k` arrays

rainwoodman:
The nditer interface of numpy may give some additional inspirations. An iterator for ParticleMesh object? I have those in FastPM that do seem to be quite useful (PMKIter and PMXIter): https://github.com/rainwoodman/fastpm/blob/master/libfastpm/transfer.c#L118

nickhand:
Yes something like that could be useful -- I'll merge this and open a new PR

PR #208, opened 2016-07-10
cleaning up measurestats with mesh iterator

nickhand:
* this introduces a mesh iterator that can apply a function slab-by-slab, which should help the readability of the code

### DIFF
--- a/nbodykit/measurestats.py
+++ b/nbodykit/measurestats.py
@@ -4,65 +4,62 @@ import logging
 import time
 
 from nbodykit.extensionpoints import Painter, Transfer
-from nbodykit.utils.meshtools import MeshWorker
+from nbodykit.utils.meshtools import SlabIterator
 
 logger = logging.getLogger('measurestats')
 
-def timer(start, end):
-    """
-    Function to format a start and end time, as returned
-    by :func:`time.time`
-    """
-    hours, rem = divmod(end-start, 3600)
-    minutes, seconds = divmod(rem, 60)
-    return "{:0>2}:{:0>2}:{:05.2f}".format(int(hours),int(minutes),seconds)
-
+#------------------------------------------------------------------------------
+# Fourier space statistics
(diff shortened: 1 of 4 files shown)


nickhand:
@rainwoodman -- take a look at the new [project_to_basis()](https://github.com/nickhand/nbodykit/blob/ac0f4a0759952bca6934895f878abeae91c31f7e/nbodykit/measurestats.py#L517). I think this does a pretty good job hiding the details but leaving the general algorithm

rainwoodman:
Indeed. I am slightly in more favor of 

```
for slab in SlabIterator(x3d, axes=0):
    xsq = slab.normp(p=2) # distance, length, ...
    # or insist on x = slab.norm(p=2)
    ....
    ....

```
this way we can avoid defining a temporary object and a temporary function.
The SlabIterator can be a generator function that yields Slab objects. p=2 is euclidean distance.

Yu

rainwoodman:
I do feels like the singular can go into the worker/iterator object too, by passing `symmetric` into the constructor.

nickhand:
that seems reasonable to me -- I think I could go either way. The temporary function is a bit ugly, but I did manage to hide the for loop though :)

rainwoodman:
Yes. but the cost is two temporary names. In addition,I have a rather strong opinion against 'worker's... I guess it is because the name is overly generic.

nickhand:
Okay, I've got a new version of the SlabIterator running -- I've confirmed it works with generic axis values now. I've also changed the syntax regarding the "singular plane" in hopes of making it a bit more transparent. Let me know what you think.

I just went with a simple `norm2` that returns the square of the norm on the slab, for the moment. I think that's the main use case at the moment. I'm happy for you to implement a more general p-norm (or any other small changes) after I merge.

rainwoodman:
It looks pretty cool indeed.

Issue #197, opened 2016-06-18
theory-based DataSource mocks + power spectrum dependency


nickhand:
The code for the Gaussian, lognormal + Zel'dovich mocks themselves has been done for awhile -- the issue is that they all start from an input linear power spectrum. The user could pass this in to the DataSource, but the cleanest solution is to use our existing cosmology framework and generate it from that. 

I think I can actually extract a basic version of my python wrapper of CLASS to do this; it would be a useful community tool either way. We could then include this external package like an optional dependency for certain DataSource classes, similar to halotools. 

If I did this, I should probably make the CLASS wrapper interface well with the astropy cosmology class, since it could be more broadly used, i.e., something like an astropy-affiliated package. 

Thoughts, @rainwoodman?

nickhand:
Also, generation of direct Gaussian fields in Fourier space would be another use of GridDataSource supporting Fourier modes, i.e., #100

nickhand:
close this for now, since most of this is addressed by #209

PR #209, opened 2016-07-12
introducing a Zeldovich sim data source 

nickhand:


nbodykit/plugins/datasource/ZeldovichSim.py
```diff
@@ -0,0 +1,96 @@
+from nbodykit.extensionpoints import DataSource
+from nbodykit import utils
+
+import numpy
+import logging
+
+logger = logging.getLogger('ZeldovichSim')
+         
+class ZeldovichSimDataSource(DataSource):
```
rainwoodman: I wonder if it is the time to add a FastPM python binding such that we can use some of the routines there, for 'sim' datasets.  e.g. the routine to fill the linear gaussian field with gadget random seeds could be useful here. Then we can also avoid reading the simulation from disk etc.

rainwoodman:
I am guessing conda comes with a ready-to-use fortran compiler that may work better than the ubuntu one. Does class require fortran?

nickhand:
victory! 

A lot of the issues were coming from the `classylss` side of things. basically, we just need to specify CXX and F90 to their GNU compiler versions, i.e., g++ and gfortran.

The fortran dependency comes from including the source code for FFTLog, which is used for power spectrum <--> correlation functions

nbodykit/plugins/datasource/ZeldovichSim.py
nickhand: that's an interesting idea...we'll certainly need to start increasing the synergy between FastPM and nbodykit once we get going on the emulator a bit more, so this is definitely worth thinking about. How much work would it be?

rainwoodman: I crafted a 'proper' C++ binding in around 16 hours. It wouldn't be too much more difficult to do something like this with Cython.

rainwoodman:
we need "matching" cmpilers with classlss with class and fft log?

Does this mean we need class and fftlog installed before pip install classlss?

nickhand:
The CLASS compilation is independent -- `classylss` downloads CLASS and uses the built-in Makefile. The default use `gcc` to compile the `.c` files of CLASS to object files. 

Then `classylss`  uses the `gfortran` to compile the `.f` files of fftlog to object files, and uses the C++ compiler `g++` to compile the main library of `classylss`. Then all of the object files are linked together into a library.

I'm by no means an expert here, so let me know if we should be doing something else

rainwoodman:
I see. I forgot classlss is CPP!

nickhand:
okay I am going to merge this and close #197

Issue #214, opened 2016-07-20
Case Sensitive


rainwoodman:
Nick could you take a look how to make the config parser case insensitive on 'names'? I couldn't find a place to start.

nickhand:
Let me see if I understand what you want to do. You want to basically call `lower()` on the dictionary keys returned by the YAML parser and the names in the schema before checking if the name is present?

nickhand:
closed by #216

PR #216, opened 2016-07-20
case insensitive parsing of configuration file names

nickhand:
* implements case insensitive parsing of names when comparing names in the Schema and configuration file 
* adds some more documentation

Is that what you had in mind @rainwoodman ?

### DIFF
--- a/nbodykit/utils/config.py
+++ b/nbodykit/utils/config.py
@@ -68,13 +68,10 @@ def fill_namespace(ns, arg, config, missing):
     # the name of the parameter
     name = arg.name.split('.')[-1]
     
-    # names of parameters present in parsed config (lowered)
-    config_names = [k.lower() for k in config]
-    
     # no subfields
     if not len(arg.subfields): 
-         
-        if config is not None and name.lower() in config_names:
+    
+        if config is not None and name.lower() in [k.lower() for k in config]:
             value = config.pop(name)
             try:
                 setattr(ns, name, ConstructorSchema.cast(arg, value))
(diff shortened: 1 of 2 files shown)


rainwoodman:
Looks about right, but which name is from the file and which name is for the schema?

nbodykit/utils/config.py
```diff
@@ -65,31 +65,32 @@ def fill_namespace(ns, arg, config, missing):
         a list to add arguments that are missing, i.e., required and not
         present in the input configuration files
     """
-    # the name of the parameter
-    name = arg.name.split('.')[-1]
+    # the name of the parameter (as taken from the schema)
+    schema_name = arg.name.split('.')[-1]
```
nickhand: this is the name of the argument taken from the schema

nbodykit/utils/config.py
```diff
@@ -65,31 +65,32 @@ def fill_namespace(ns, arg, config, missing):
...
     
     # no subfields
     if not len(arg.subfields): 
     
-        if config is not None and name.lower() in [k.lower() for k in config]:
-            value = config.pop(name)
+        # check if the schema argument is present in configuration file
+        if config is not None and schema_name.lower() in [k.lower() for k in config]:
```
nickhand: and the keys of the ``config`` dict here are the names of the parameters from the configuration file

nbodykit/utils/config.py
```diff
@@ -30,31 +38,59 @@ def construct_mapping(loader, node):
...
     
-    if not len(arg.subfields):  
-        if config is not None and name in config:
-            value = config.pop(name)
+    # no subfields
+    if not len(arg.subfields): 
+    
+        # check if the schema argument is present in configuration file
+        if config is not None and schema_name.lower() in [k.lower() for k in config]:
+            value = config.pop(schema_name)
```
rainwoodman: Will this fail if the cases don't match.

nickhand: ah yes, yes it will

nickhand:
okay to close #214 ?

PR #217, opened 2016-07-20
Paintgridalg

rainwoodman:


nickhand:
this looks good -- do we want to make the GridDataSource read the output of this algorithm?

rainwoodman:
I am working on a refactor. There is currently a GridSource extension point. And a BigFileGridSource plugin that can be plugged in. 

I modified the PaintGrid algorithm to allow it use GridSource input. Maybe shall do that for PeriodicBox as well. Any other place that a GridSource is used?

Shall we deprecate GridPainter and the GridDataSource plugin?

rainwoodman:
Do you have old grid files that you plan to use?

nickhand:
No, we can get rid of the old GridDataSource and GridPainter. I don't even think GridDataSource was functional. 

I agree that it would be good to modify PeriodicBox and any other places that the new grid functionality can be used easily

rainwoodman:
Sure. I hacked DefaultPainter, so it will accept both GridSource and DataSource. I'll clean up the unused painters.

PR #220, opened 2016-08-03
adds a test module for FKPCatalog and Weight/FKPWeight breakdown

nickhand:
two new additions: 
* test module for FKPCatalog for all of the various input scenarios
* FKPCatalog now can handle FKPWeight and Weight


nickhand:
this also closes #220

PR #221, opened 2016-08-05
WIP: Improving the Grid to support reading k and Fourier resampling

rainwoodman:


rainwoodman:
It will need the new version of mpsort (0.1.8) that I have just tagged.

rainwoodman:
I liked normalize, frho and fk functions in bf209a9 . Could it be an easier replacement than those giant ugly transfer functions?

PR #222, opened 2016-08-09
WIP: dask + file formats implementation

nickhand:
we'll use this PR to explore using dask and using file formats


nickhand:
@rainwoodman -- take a look at this new potential syntax for "file formats" and dask interface

A few initial thoughts: 

* I think the dask array is the right choice over data frames:
     - The lack of vector data types is really ugly to get around 
     - The interface to the DataFrame is just not as intuitive as the array syntax -- I've used pandas a lot and interacting with DataFrames still isn't easy as simple arrays
     - Using dask arrays is more in line with our current interface, where the "read" functions in different datasources are all operating on  structured arrays
* I think one existing problem is that the DataSource right now is handling both the reading of the file and the transformation of what is read, which are really two different tasks 
* I think we want several builtin FileFormat classes like this BinaryFile class and then maybe the DataSource should specify which file format(s) it can read? 
* It think it should be possible to define a fixed interface for different file types (binary, h4py, csv, bigfile) and functions to compute them to dask arrays

nbodykit/dask/test_tpm.py
```diff
@@ -0,0 +1,75 @@
...
+        s = cls.schema
+        s.description = "read snapshot files from Martin White's TPM"
+        s.add_argument("path", type=str, help="the file path to load the data from")
+        s.add_argument("BoxSize", type=cls.BoxSizeParser,
+            help="the size of the isotropic box, or the sizes of the 3 box dimensions")
+        s.add_argument("rsd", choices="xyz", help="direction to do redshift distortion")
+        s.add_argument("bunchsize", type=int, help="number of particles to read per rank in a bunch")
+
+    
+    def Position(self, data):
```
rainwoodman: You sure it is OK to mix the names of columns and methods?

nbodykit/dask/test_tpm.py
```diff
@@ -0,0 +1,75 @@
...
+    
+    def __init__(self, path, BoxSize, rsd=None, bunchsize=4*1024*1024):
+        
+        self.path = path
+        self.BoxSize = BoxSize
+        self.rsd = rsd
+        self.bunchsize = bunchsize
+        
+        dtypes = [('Position', ('f4', 3)), ('Velocity', ('f4', 3)), ('ID', 'u8')]
+        self.storage = BinaryFile(self.path, dtypes, header_size=28)
```
rainwoodman: For example, GADGET uses F77 unformatted blocks, then you need a 'gap_size' parameter at least. 

Talking about GADGET it is far more complicated. It may be possible to peek the header to tell the schema, e.g. if the Position is in f8, if the ID is in u4. In Gadget these are stored as attributes in the header file.

It gets even more complicated when we allow multiple particle types. The column length is variable in this case! For example,  position contains all data types, but gas temperature only contains the gas particles. Gadget also has a stupid optimization that he mass of dark matter particles is sometimes not stored in the freaking files, but as a single number in the header block, while all other particles are stored in the mass column!

nbodykit/dask/test_tpm.py
```diff
@@ -0,0 +1,75 @@
...
+        as dask-delayed functions
+        """
+        if self.rsd is not None and 'Velocity' not in columns:
+            columns.append('Velocity')
+        
+        # these are the column functions, delayed so they can execute together
+        tasks = [delayed(getattr(self, col)) for col in columns]
+        
+        # loop over chunks of the file storage
+        for data in FileIterator(self.storage, columns, chunksize=self.bunchsize, comm=self.comm):
```
rainwoodman: FileIterator yields partition, not data?

nbodykit/dask/test_tpm.py
```diff
@@ -0,0 +1,75 @@
...
+        return data['Velocity'] * self.BoxSize
+        
+    
+    def read(self, columns, full=False):
+        """ 
+        Return the columns of this DataSource, by executing them
+        as dask-delayed functions
+        """
+        if self.rsd is not None and 'Velocity' not in columns:
+            columns.append('Velocity')
```
rainwoodman: We will get one extra Velocity column. I guess we can worry about this later.

nbodykit/dask/test_tpm.py
```diff
@@ -0,0 +1,75 @@
...
+        
+        # these are the column functions, delayed so they can execute together
+        tasks = [delayed(getattr(self, col)) for col in columns]
```
rainwoodman: delayed_columns = [ delayed( ...) for col in columns ]?
The we yield compute on the delayed columns?

But if we immediately compute the delayed object, what is the point of using delayed?

Maybe each column named attribute shall be made into a delayed, and we just return a list of the delayed objects. Then the down stream (e.g. painter) can compute them as needed?

rainwoodman:
I am leaning towards datasource/FileTypePlugin returns a list/dict of list of dask.arrays / dask.delayed for all columns, normalized to a given set of units (partitions scheduled to this rank). Then we will have a Transformer that does RSD, RA-DEC->XYZ, and other renaming stuff, returns a dict of list of daskarrays / dask.delayed.

A painter will take a transformer's output dictionary, pick the columns according to the painter's documented protocol, compute them, and calls ParticleMesh to paint them.

rainwoodman:
We can call Transformer PointDataSource, and the first object that normalizes files FileTypePlugin?
Does this proposal look sane?

nickhand:
Yes, I am thinking along the same lines...something to read and return data as read from file and then something to transform it. 

A few thoughts:
* I think we should be able to pass along delayed objects. When reading is expensive, you want to use `dask.compute(*delayed_columns)` to compute all columns at once, so just keep that in mind
* I think the "columns as delayed functions" was the easiest interface with dask...dask array is still limited for our purposes,  (you can't mutate a dask array), but the delayed feature works nicely
* We should think about how things like ZeldovichSim and UniformBox fit into this. Perhaps we use a FileSource and SimulationSource and the transformation can handle any "Source" of data?

nbodykit/dask/test_tpm.py
nickhand: I think this would work...I was imagining decorating the column functions with "@Column" that handles applying "delayed" to class functions and maybe registering them?

The thing to keep in mind is the all the delayed functions should be computed at the same time...but this can be done at any point in the pipeline before pmesh I think

nbodykit/dask/test_tpm.py
nickhand: At this point, it is yielding a specific slice of a dask.array. I need to fix it so the chunks align with the number of ranks. If you aren't careful and you slice the dask array that isn't chunked properly, the rank has to read the whole (probably way too large) array and then return the slice.

nbodykit/dask/test_tpm.py
nickhand: I am not sure, but for the moment, I couldn't think of anything

nbodykit/dask/test_tpm.py
nickhand: Yes, I was looking at the existing stuff for Gadget and quickly realized my format was overly simplified. I'll stop by so we can chat more about what we want to do here...I was looking at how yt, etc handle Gadget, and I think at least found some data to test on.

nickhand:
The sliceable file format straight to dask array is very very nice. I've got it working for [csv files](https://github.com/nickhand/nbodykit/blob/fa933cae39f9fbb964b07f59f2470a439f664624/nbodykit/dask/csv_file.py#L152) and [binary files](https://github.com/bccp/nbodykit/blob/fa933cae39f9fbb964b07f59f2470a439f664624/nbodykit/dask/binary_file.py#L223)

Should be relatively easy to add HDF5, Bigfile, and FITS file formats too and then we have a pretty good representation. HDF5 and Bigfile already can be converted straight to dask arrays

nickhand:
Also I was brainstorming names for the module to holding the file formats. I kind of like `nbodykit.hermes` -- the idea being something like the module passing data quickly from disk to the algorithms...

rainwoodman:
Does it make sense to have hermes a separate package that nbodykit depends
on?

The name is already taken though.

nickhand:
closing this in favor of #225

PR #223, opened 2016-08-12
WIP: replace pmesh.particlemesh with pmesh.pm.

rainwoodman:
pfff.


rainwoodman:
@nickhand 

I had to move paintbrush to a parameter of DefaultPainter. The new logic is PeriodicBox sets paintbrush if it is not set in DefaultPainter.

rainwoodman:
@nickhand I still have multiple failures in FKP. Do you want me to fiddle with it? I do need you to look at the Pandas HDF5 failure. There are so many nans in the 'correct' file that I am a bit suspicious .... Did I mess up Ntot?

nickhand:
@rainwoodman I'll take a look now

nickhand:
the pmesh normalization has changed? or is this a [bug](https://github.com/rainwoodman/pmesh/blob/master/pmesh/pm.py#L220)

Seems like we went from r2c dividing by Nmesh^3 to now r2c divides by only Nmesh

nickhand:
never mind -- you are taking the product .

my changes to Bianchi implementation results in orders of magnitude difference, so I thought this might be the fix

rainwoodman:
What changes?

nickhand:
about to open a PR on your newpm branch

I think I have all of the FKP related stuff working -- difference was that pmesh paints number now, not number density

PR #227, opened 2016-08-15
WIP: interlacing with new pm.

rainwoodman:
1. add a helper method in Painter for painting shifted positions.
2. DefaultPainter takes 'interlaced' argument
3. If we can move the transforms out of DefaultPainter, then we can define a separate InterlacedPainter. 
We may want to merge/take a closer look at this after the io PR is resolved, since files have moved and this change is rather trivial.


nbodykit/core/painter/DefaultPainter.py
```diff
@@ -68,19 +69,19 @@ def paint(self, pm, datasource):
...
                             self.basepaint(real, position, weight=weight, paintbrush=self.paintbrush)
                         else:
-                            self.interlaced_paint(real, real2, position, weight=weight, paintbrush=self.paintbrush)
+                            self.shiftedpaint(real, real2, position, weight=weight, paintbrush=self.paintbrush)
                         Nlocal += len(position)
             c1 = real.r2c()
             c2 = real2.r2c()
-            H = self.pm.BoxSize / self.pm.Nmesh
 
+            H = pm.BoxSize / pm.Nmesh
```
rainwoodman: This is a bit suspicious to me. @nickhand does this look right?
It appears to be
```
s1 += s2 * 0.5 * exp(I * k * (L / 2N))
```

rainwoodman: There seems to be some kind of normalization issue. Divide by sqrt(2)?

rainwoodman:
The test case is still missing. @nickhand will look at this after merging.

Issue #234, opened 2016-08-18
Integrate Datasource with Transformation and FileType 


nickhand:
following up on what was started in #225

PR #238, opened 2016-08-19
WIP: Multipart DataSource.

rainwoodman:
This will aid the transition to nbodykit.io.


rainwoodman:
Will probably address #234 here.

rainwoodman:
@nickhand Can I have a dictionary type in a schema? Will the config object fail?

nickhand:
you want to specify the dict as the type? I think it should work, but it might not doing anything if the already parsed the field to a dict. The `type` function will get called on whatever comes out of the YAML parser

rainwoodman:
I can't find a wayto pass dtype to BinaryFile. I tried building a dtype argument but because YAML uses lists the cast to numpy.dtype will fail.
Maybe we shall subcalss BinaryFile to TPMBinaryFile and GadgetBinaryFile etc as the actual FileType plugins?

rainwoodman:
So, UniformBox and ZelDovich are FileTypes in the new framework?

rainwoodman:
It doesn't feel right that way.

nickhand:
hmm, I agree it doesn't feel right that way, but maybe that is just a naming issue? In practice, they are still providing a series of columns to the user than can be transformed if desired. 

This gets back to the idea of "algorithms as DataSources" a little bit...both of these cases are really running simple algorithms to produce data. Maybe we want to expand on that keeping in mind FOFGroups, etc. 

I'll be back in town next week and we can discuss in person, and also discuss the latest on the dask front

nickhand:
replaced by #245

PR #242, opened 2016-08-19
remove `registry` from plugins; handle everything through plugin_manager

nickhand:
I don't think the registry attribute is necessary

handling everything through plugin_manager simplifies a lot of the code

### DIFF
--- a/bin/nbkit.py
+++ b/bin/nbkit.py
@@ -5,9 +5,7 @@ import logging
 import os, sys
 from mpi4py import MPI
 
-from nbodykit import plugin_manager, algorithms
-from nbodykit.core import Algorithm
-from nbodykit.core import DataSource, Transfer, Painter
+from nbodykit import plugin_manager
 from nbodykit.plugins import ListPluginsAction, EmptyConfigurationError
 from nbodykit.plugins.fromfile import ReadConfigFile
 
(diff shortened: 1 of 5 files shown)


nickhand:
@rainwoodman does this look reasonable to you?

rainwoodman:
It looks alright. This reminds me the issue that in a plugin module we can't import plugin_manager from top-level, since when a plugin module is executed the plugin_manager is not yet exposed to the global namespace of plugins. Is there a way to fix that? Or do we want to fix that at all?

nickhand:
that's a good point. I'm not sure that I mind. We can do something like: 

`from nbodykit.plugins.manager import PluginManager`

at the top of modules and the `plugin_manager = PluginManager.get()`

if so desired

rainwoodman:
My concern is that if the plugin_manager object is not properly initialized yet, we probably shall not be able to obtain its reference. So I sort of felt the old behavior captured some correct things but in an awkward way. 

The new behavior is confusing. Maybe rename the old get to create? Usually the singleton 'get' function doesn't take any arguments, does it? Because the argument implies the object relies on the current scope, however a singleton shall not.

nickhand:
Sure, that makes sense. At least from the user side, the plugin manager will always be there since it is initialized nbodykit/__init__.py

The old loader had the same issue, right? You couldn't reference other plugins from a different plugin module, I think

rainwoodman:
I suspect so.

In MultiFile I had to use the plugin_manager to look up the FileType plugin. As long as I import plugin_manager in side the plugin class member function, we are fine -- Since code there are supposed to be invoked after plugin_manager is already in a valid state.

nickhand:
Yeah I'm not sure there is a simple or even robust way of resolving such an import at the top of the file. The `get()` method might be the best we could do. I thought the singleton pattern was rather intuitive for the plugin manager but feel free to disagree

rainwoodman:
It's just a name. I think what we have is fine.

Two questions to think about though:

Why do we ensure there is only plugin_manager per python process? (Singleton)

What if we choose to ensure one plugin_manager per nbkit main function instead?

nickhand:
Hmm, that's a good point. Something to think about. How would even enforce that so that it worked in things like IPython?

I'll merge this for now

rainwoodman:
e.g. creating a NBKit object? that cares about plugins and communicator and cosmology? I think it is an overkill.

nickhand:
yes that could be overkill, at least at this point

Issue #237, opened 2016-08-18
Eliminate / simplify the Transfer Functions.


rainwoodman:
1. The Transfer functions no longer need the PM object.
2. The DefaultPainter accepts f(k) and f(rho) arguments. It replaces some transfer functions. We may want to prefer providing building blocks in the namespace where f(k) and f(rho) were ran rather than provide transfer functions as an extension point -- it is something a user would like to fiddle more and a demanding a plugin is too heavy.

rainwoodman:
deprecated in favor of v2

Issue #235, opened 2016-08-18
Update documentation / add tutorials section with handy notebooks 


nickhand:
Before completing #234, we should update the base docs (which are currently broken) to describe the new API. 

Also, we should add a tutorials section with a number of notebooks. An initial list: 
- algorithm examples, i.e., ZeldovichSim + FFTPower might be nice to plot 
- move Dataset to a notebook tutorial
- example of a user plugin
- different kernels
- interlacing comparison
- loading data with IO library + out-of-core ops with dask

I imagine some of these plots / description could go in the paper. Thoughts, @rainwoodman?

nickhand:
closing this in favor of #366

PR #244, opened 2016-08-20
Finish interlacing tests

nickhand:
It appears interlacing is working. I've added the test data to nbodykit-data, so the tests should hopefully all work now.

I've added a [notebook](https://github.com/nickhand/nbodykit/blob/finish_interlacing/docs/notebooks/interlacing.ipynb) in the docs directory showing a comparison.  For both TSC and CIC, introducing aliasing improves accuracy, and TSC with interlacing is nearly exactly to Nyquist (when compared to the result from a  2x mesh).

The existing CIC and TSC transfer corrections should probably be renamed -- they correct for both the paint window (exact) and the aliasing contribution (only exact for Poisson noise). Perhaps we can do that when we finish #237. 

I've added transfers that just correct for the window. In interlacing, I believe the idea is that interlacing can help remove the aliasing to a degree where only a correction for the painting window needs to be applied.


nickhand:
this is also the first step towards #235, with the docs/notebooks directory added

nickhand:
@rainwoodman take a look at this [notebook](https://github.com/nickhand/nbodykit/blob/finish_interlacing/docs/notebooks/interlacing.ipynb) when you get a chance and see if that makes sense

rainwoodman:
It is very inspiring. Would you also take a look at the cross correlation coef between truth and various fields? You can use the resample method of the field to downsample the truth to a lowresolution output.

PR #259, opened 2016-10-31
Avoid numpy bugs due to stealing dtypes.

rainwoodman:


rainwoodman:
Travis was green so I am merging this. It should alleviate some strange core dumps on nersc.

PR #265, opened 2016-12-05
initial commit with 2 examples of possibly workflow in "dev"

nickhand:
* follows the "from nbodykit.lab import *" syntax
* uses a Zeldovich particle source
* only algorithm added is FFTPower, but seems to be working
* example in regular mode and "batch" mode, which iterates over tasks nicely


nickhand:
@rainwoodman take a look at the `dev` directory for two possible examples of the new workflow. lots of design-related stuff to discuss, but I am really liking this new workflow

rainwoodman:
I'll merge this so I can play with it.

PR #272, opened 2016-12-09
WIP: cleaning up io library and CSV file

nickhand:


nbodykit/tests/test_lab.py
```diff
@@ -51,7 +48,7 @@ def test_paint(comm):
     output = "./test_paint-%d.bigfile" % comm.size
     alg.result.save(output)
 
-@MPIWorld(NTask=[2, 3, 4])
+@MPIWorld(NTask=[2, 3, 4], required=2)
```
nickhand: @rainwoodman -- without ``required=2`` this test appears to get run with NTask=1, must be a bug in mpi4py_test

nbodykit/io/csv.py
```diff
@@ -1,17 +1,136 @@
 import numpy
-
+import os
+from pandas import read_csv 
+    
 from ..extern.six import string_types
 from . import FileType, tools
 
+class CSVPartition(object):
```
nickhand: I'm using this class to cache the partitions through the ``value`` attribute, which calls read_csv on the string of bytes when requested

nbodykit/io/csv.py
```diff
@@ -1,17 +1,136 @@
...
+    
+    Returns
+    -------
+    partitions : list of CSVPartition
+        list of objects storing the data content of each file partition,
+        stored as a bytestring
+    sizes : list of int
+        the list of the number of rows in each partition
+    """
+    from dask.bytes.utils import read_block
```
nickhand: @rainwoodman this is the only dependency from dask still...it handles partitioning the file into blocks, always ending on the newline character. This function seems pretty stable to me...

rainwoodman:
Merge?

Issue #249, opened 2016-09-13
Port/Invent Gaussian/Linear source to GridSource


rainwoodman:

nickhand:
closed via #274

PR #274, opened 2016-12-12
adds HDF support and Grid sources

nickhand:
* test case added for HDF file
* test cases added for linear grid and bigfile grid

addresses #249


nickhand:
@rainwoodman shall I merge this?

PR #275, opened 2016-12-13
WIP: Notransform

rainwoodman:
Remove transforms from painter. 

Add a Field object that allows more flexible 'in-place' transformations.


rainwoodman:
@nickhand I am thinking FFTPower shall expect Field objects as inputs.

We can add more actions to Field objects to correct for window effects. These actions are deferred and not ran till Field.paint is called.

rainwoodman:
@nickhand  It is to the point where I think I need to touch fftpower. There are two possibilities:

1. FieldStudio creates ComplexField / RealField that goes into FFTPower;
2. FFTPower takes 1 or 2 FieldStudio objects along with 1 or 2 Source objects.

I lean towards 1. 2 has too many arguments.

There is a slight glitch with FieldStudio. Source has a boxsize (for normalization) and FieldStudio has another boxsize (the size of the studio). They need not to match but they shall usually match. There is no way to capture this logic automatically -- FieldStudio is created without knowing a source. 

Field created by FieldStudio carries an attr member that has all attrs of the source plus shotnoise, this shall be sufficient for the FFTPower algorithm to infer the compensation  -- or we can put compensation as a method in FieldStudio, which we call before passing the RealField/ComplexField to FFTPower.

nickhand:
Interesting. If I understand the logic correctly, FieldStudio is basically the particle mesh + transformations to the field?

I think I was imagining the old Painter as something like FieldStudio. What is the advantage of having to pass the source to the FieldStudio, rather than setting it as an attribute of the Source? I could imagine the Source having functions "to_mesh" (i.e, the same as the current paint for Sources) and  "from_studio" which is the current FieldStudio.paint I think.

rainwoodman:
- I think FieldStudio is Painter. but the time I wrote this the Painter name was occupied.

- If FieldStudio is pushed into Source, then we push ParticleMesh into the source. I don't see much of a problem. It actually solves the glitch about BoxSize too.

rainwoodman:
This way FFTPower will not create a ParticleMesh object.

nickhand:
Yes, I think this makes the most sense to me

rainwoodman:
Source shall have a ParticleMesh (dtype, BoxSize, Nmesh).

- abstract methods to_real_field() and to_complex_field() that shall be overriden by subclasses
- an interface (r2c, c2r, apply, paint) for painting a transformed field
- Field has added .attrs attribute and a .save() member that saves to bigfile.

ParticleSource is a subclass of Source, with
- 'interlaced' and 'window' attributes for painting
- '__getitem__' for IO
- override to_real_field()

FFTPower takes the pmesh Field objects as input. r2c automatically.
 
I'll draft this up after lunch.

nickhand:
this sounds good to me

rainwoodman:
I removed set_transfer from FFTPower. Now the compensation for window is directly built in source.

And you can apply new transformations to the field in real or complex with the (somewhat cryptic) apply method.

rainwoodman:
I also changed los in FFTPower to a vector.

nickhand:
this is looking really good. I like it

nbodykit/base/particles.py
```diff
@@ -82,34 +89,41 @@ def set_transform(self, *transform, **kwargs):
...
-        """
-        self._painter = painter
-        
+    @property
+    def interlaced(self):
+        return self.attrs['interlaced']
+
+    @interlaced.setter
+    def interlaced(self, interlaced):
+        self.attrs['interlaced'] = False
```
nickhand: this is a bug @rainwoodman

rainwoodman:
OK. I am done with the big changes.

The 'u' looks alarming: 
https://travis-ci.org/bccp/nbodykit/jobs/183799425#L560

Take a look if there are other issues?

nickhand:
fixed the string issue, reviewing the rest now...

nbodykit/base/particles.py
```diff
@@ -15,22 +19,30 @@ class ParticleSource(object):
...
-        self._painter = Painter()
+        GridSource.__init__(self, BoxSize=BoxSize, Nmesh=Nmesh, dtype=dtype, comm=comm)
 
+        self.attrs['compensated'] = True
+        self.attrs['interlaced'] = False
+        self.attrs['window'] = 'cic'
+
+        # if size is already computed update csize
+        # otherwise the subclass shall call update_csize explicitly.
+        if self.size is not NotImplemented:
```
nickhand: ah, I see the logic in Zel'dovich

nickhand:
I think this is a big improvement...There is a lot more functionality in the "lab" framework now 

"DumpField" is broken and we can remove it, right? The "test_paint" case is the cleaner way to do this. I think we benefit from loosening up the rigid "Everything is an Algorithm" framework

rainwoodman:
I think we can get rid of the transforms as well. Since `__getitem__` returns dask arrays, we can always directly create new columns / override columns in the source.

rainwoodman:
We need a way to serialize cosmology into a bigfile.

Can we insist .attr only contains array_like elements?

nickhand:
I'm just about done with the Eisenstein and Hu power class and then we can remove classylss. Annoyingly, astropy's cosmology isn't serializable to a dict at the moment, so we'll still need to some sort of conversion function. But we will need to store `n_s` and `sigma8` values anyways...

The classylss cosmo dict returns strings because the CLASS code takes string parameters, and with mixed-type parameters, this was the easiest to do...

rainwoodman:
I added a converter in utils.py that uses classylss to cast values to a float. If we are to remove classylss from the core dependency may need to port that function over.

rainwoodman:
In any case, shall we merge this for now?

It still feels strange that we rely on column reassignment to determine which column the data source uses as position, weight and selection.

Maybe we really want a ParticleSource that doesn't implement the GridSource interface, but a special GridSource that adapts a ParticleSource, which takes column names.

nickhand:
I think handling the selection using boolean (dask array) slices of ParticleSource would be more intuitive and elegant. We should be able to do that in a similar way to how ndarray handles slices, i.e., something like a `base` attribute

I agree it would be nice to remove the fixed "Position" and "Weight "columns...need to think about how to best do that

rainwoodman:
What about
```
GridFromParticle(position=particles['Position'], weight=particles['Velocity'][:, 0], selection=particles['Mass'] > 1e10)
```

nickhand:
Yes probably something like that. It feels like we are still missing something. Perhaps "Grid" should really be a "Field" and we either generate "Fields" by:
1. loading them from arrays on disk
2. specifying position, weight, selection from a ParticleSource
3. calling paint on a ParticleSource, with the specified position, weight columns

Also, does the current logic allow for compensating painted grids that are loaded from disk? Say we paint a particle source, save the grid, and then want to do FFTPower later on? 

Are you around today?

rainwoodman:
Yes. I am on campus. 

I almost vetoed 3) during lunch. There shall be only one way of doing one thing. The protocol is Position Weight Selection then we stick to it. 
Opening up any columns introduces a confunsion between column name and column. 

The name Field has already been take by ParticleMesh, so it is a bit confusing to have another 'Field' which is neither RealField or ComplexField.

nickhand:
ok I'll come down from the lab to chat

PR #282, opened 2016-12-19
adds cosmology and pmesh random field generation

nickhand:
* no more classylss, in favor of Eisenstein and Hu power spectra
* `nbodykit.cosmology.Cosmology` is dict-like and adds growth function and growth rate calculations


nickhand:
@rainwoodman one thing I am not sure about is the random seed in parallel used in `poisson_sample_to_points`. The density field in each cell is already random, and then we draw a value from the Poisson distribution with that mean. Can we use the same global seed across all ranks to do that? It seems potentially problematic...

nickhand:
hmm, @rainwoodman, what do you think the failed test in test_zeldovich_velocity

https://travis-ci.org/bccp/nbodykit/jobs/185288936#L780

Is this just round-off error and that weighting by Velocity means that the mean will be close to zero?

rainwoodman:
- possion_sample_to_points. What about to converting the gaussian field (real space) to a uniform distribution, then use that to seed a Possion?

rainwoodman:
- error : I think it will go away if you switch to 'f8'. That's how I ended up loosing 'f4' rtol to 1e-5.

nbodykit/cosmology/ehpower.py
```diff
@@ -0,0 +1,313 @@
...
+        def integrand(lnk):
+            k = np.exp(lnk)
+            kr = k*r
+            j1_kr = np.sin(kr)/kr**3 - np.cos(kr)/kr**2
+            T = self.transfer(k)
+            return k*k*k * (k ** self.cosmo.n_s) * (T ** 2) * (3*j1_kr) ** 2
+            
+        norm = self._norm * (self.sigma8 / self._sigma8_0)**2
+        lnk = np.log(np.logspace(-4, 4, 4096))
+        return self._norm / (2*np.pi**2) * simps(integrand(lnk), x=lnk)
```
rainwoodman: Yin may disagree about using simps here. We'll see what he thinks.

nickhand:
are you familiar with the `randomstate` (https://pypi.python.org/pypi/randomstate) package at all? It seems to have parallel stream generation for our purposes and be fairly advanced. It looks like there is an open issue for incorporating it into numpy at some point

nickhand:
see http://bashtage.github.io/ng-numpy-randomstate/doc/build/html/pcg64.html

nickhand:
re: poisson sample are you suggesting generating another random field in Fourier space, r2c, and then converting that? It seems like it might just be simpler to have root generate the random number and then scatter accordingly. We need the Poisson randoms as well as the 3 uniform randoms for offsets within the cell size

nickhand:
okay, at least with 655533f, the random numbers are generated coherently now. we can see if we can think of cleverer things to do in the meantime

rainwoodman:
Yes. The rngs in numpy doesn't allow vectorized seeds. I think we may be able to put something in PMesh that take the seed and a mean number count field, sampling it to a integer field by poisson. 

The poisson sampling code in numpy is quite simple (there are two functions for large and small lambda), and we can feed each cell a different random sequence -- which are uncorrelated. 

Can we merge this, such that I can start working on a FastPMParticles source?

rainwoodman:
Merging because we need this to run it on Jupyter.nersc.gov. classylss won't compile due to dependency on swig...

nickhand:
yes sounds good....was in the car traveling all day

rainwoodman:
Well. What's the plan about swig with classylss though? Maybe you can ship a few more files such that swig doesn't have to be ran? Or just require installing swig (NERSC anaconda doesn't have it by default)

nickhand:
I had a version that downloaded swig and installed it if it wasn't found, which was relatively painless. I should be able to do that. 

Did you still want to remove the classylss dependency in v2.0?

rainwoodman:
I think that's not a good approach. What about including gcl.py and gcl_wrap.cpp in sdist?

nickhand:
Okay, I've included those files, but annoyingly jupyter.nersc is using g++ version 4.4, which is old enough that it doesn't support some of the newer c++11 features. In particular, it doesn't support std::bind, which the integration classes rely on pretty heavily. I don't see an easy way to fix this...they are supposedly upgrading jupyter.nersc soon, right?

rainwoodman:
The anaconda one shall be pretty recent though.. Is classylss built against the system g++?

rainwoodman:
Bummer! jupyterhub doen't have anaconda gcc.

rainwoodman:
Edison uses gcc 6.1.0. Jupyterhub uses gcc 4.4. I think we shall file a bug to NERSC help desk.

nickhand:
yes I think that makes sense...do you want to file that?

Issue #283, opened 2016-12-20
HDF reader is very very slow.


rainwoodman:
I created a source like this
```
source = Source.File(HDFFile, '00000400-40-wig/fof_1.0000.hdf5',
                     BoxSize=1380.0, 
                     M0 = 27.75e10 * 0.307 * 1380**3 / 2048 ** 3)
# cleaning
source['Position'] = source['Position'] * source.attrs['BoxSize']
source['Velocity'] = source['Velocity'] * source.attrs['BoxSize']
source['Mass'] = source['Length'] * source.attrs['M0']
source['Selection'] = source['Mass'] > 1e13
```

Then either computing a column or running FFTPower becomes extremely slow. The hdf5 file can be accessed from my home directory , cosmosim/vacan/00000400-40-wig/fof_1.0000.hdf5. (or any old nbodykit fof catalogues).

rainwoodman:
@nickhand  Where does it smell bad in the reader?

nickhand:
How large are the catalogs? I don't have access to cosmosim. If you push to an open scratch dir or m779 I can take a closer look. 

I wonder if there are many dask chunks, then the calculation of each dask chunk might be reading the full FOF catalog then slicing it...

nickhand:
or maybe this is tied to h5py's version of fancy indexing? i.e., http://stackoverflow.com/questions/21766145/h5py-correct-way-to-slice-array-datasets

rainwoodman:
No I don't think so. We select after obtaining the numpy arrays.

You can look at any hdf5 file from ~yfeng1/desi/baoshift/run4/*/*.hdf5 . 
Worse case we can cache these like csvfiles. I'll take a look if converting bigfile we suffer from slow down too.

rainwoodman:
I can confirm bigfile reader is not as slow.

rainwoodman:
closed by #284 284

PR #284, opened 2016-12-22
Compile and batch dataset operations with hdf5.

rainwoodman:


rainwoodman:
I tested on NERSC with the old fof catalogue -- this fixes issue #283

Issue #288, opened 2017-01-04
Allow lower dimension inputs to mockmaker and ZeldovichParticles


rainwoodman:
I added support to 1d and 2d and unitary(remove_variance) noise to pmesh.

I tried to remove the requirement of 3d ParticleMesh in mockmaker.py, but I got stuck at the possion realization part. 

@nickhand take a look?

rainwoodman:
2D is good enough for now since pfft can't seem to handle 1d.

PR #289, opened 2017-01-04
addresses Poisson sampling in N-dimensions 

nickhand:
addresses #288, although not completely. SlabIterator in its current form should only be used in 3D (hence, the "slab").  I think it mostly works in any dimension at the moment, but need to think about the non-singular plane and Hermitian weights. 

What did you have in mind for this, @rainwoodman ? Did you want to be able to do 1D and 2D Zel'dovich particles? Also, see rainwoodman/pmesh#17 for the case of 1D particle mesh

### DIFF
--- a/nbodykit/meshtools.py
+++ b/nbodykit/meshtools.py
@@ -54,14 +54,14 @@ class MeshSlab(object):
         Return the local shape of the mesh on this rank, as determined
         by the input coordinates array
         """
-        return [numpy.shape(self._coords[i])[i] for i in range(self.ndim)]
+        return tuple([numpy.shape(self._coords[i])[i] for i in range(self.ndim)])
     
     @property
     def shape(self):
         """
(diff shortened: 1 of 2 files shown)


rainwoodman:
- 1d ZA could be an interesting case since I recall Martin's been talking a lot about ZA being exact in 1d
- I am not sure what second order 2lpt is in 2d (does it vanish? I currently set the second order term to zero for 2d)
- The code is slightly cleaner without those hard coded [0, 1, 2]. 
- My main use case is testing the gradient etc with a 2d simulation, where things can be easily printed and investigated.
- The slab iteration in ParticleMesh works for both 2d and 3d.

nickhand:
oaky @rainwoodman, I've changed the slab iterator to use pmesh's version. I think mockmaker should work in 2D and 3D now.

shall I merge this?

PR #290, opened 2017-01-06
Generalize Subsample datasource to read all columns of the hdf5 file.…

mschmittfull:
Just a small change so the Subsample datasource reads in all columns of the hdf5 file. Previously it only read Position and Velocity but skipped ID and potential other columns, which caused a crash when trying to take subsamples of subsamples because ID was not read.


mschmittfull:
Actually something seems wrong with the subsample of the subsample. The displacement field obtained from Position-InitialPosition is screwed up when loading a subsample of a subsample file, but it looks ok when loading an old subsample generated from a full FastPM snapshot. No clue why this happens. Better don't accept the pull request for now.

rainwoodman:
The mean particle separation must have been wrong, I am guessing?

PR #291, opened 2017-01-07
WIP: adding Bianchi algorithm + cache functionality for Sources

nickhand:
in addition to Bianchi algorithm and source, biggest change is probably reliance on "cachey" + dask.cache to cache "compute" calls in Sources (if not installed, warning is printed)

### DIFF
--- a/nbodykit/base/particles.py
+++ b/nbodykit/base/particles.py
@@ -4,6 +4,9 @@ from nbodykit.transform import ConstantArray
 import abc
 import numpy
 import logging
+import warnings
+
+CACHE_SIZE = 1e9
 
 def column(name=None):
     def decorator(getter):
@@ -51,11 +54,14 @@ class ParticleSource(object):
             return da.from_array(array, chunks=100000)
 
     # called by the subclasses
-    def __init__(self, comm):
+    def __init__(self, comm, cache=False):
+        
         # ensure self.comm is set, though usually already set by the child.
-
         self.comm = comm
 
+        # initialize a cache
+        self.cache = cache
+        
         # initial dicts of overrided and fallback columns
         self._overrides = {}
         self._fallbacks = {}
@@ -65,8 +71,24 @@ class ParticleSource(object):
         if self.size is not NotImplemented:
             self.update_csize()
 
-        self._overrides = {}
-
+    @property
+    def cache(self):
+        return self._use_cache
+        
+    @cache.setter
+    def cache(self, val):
+        if val:
+            try:
+                from dask.cache import Cache
+                if not hasattr(self, '_cache'):
+                    self._cache = Cache(CACHE_SIZE)
+            except ImportError:
+                warnings.warn("caching of ParticleSource requires ``cachey`` module; turning cache off")
+        else:
+            if hasattr(self, '_cache'):
+                delattr(self, '_cache')
+        self._use_cache = val
+                
     def to_mesh(self, Nmesh=None, BoxSize=None, dtype='f4',
             interlaced=False, compensated=False, window='cic',
             weight='Weight', selection='Selection'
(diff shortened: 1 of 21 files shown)


nickhand:
still need to clean up Bianchi algorithm a bit and add a redshift histogram algorithm so N(z) column can be computed and add directly to Sources. But test case exists with the general flow in place

@rainwoodman, why don't you take a look

nbodykit/base/particles.py
```diff
@@ -44,21 +48,47 @@ class ParticleSource(object):
...
+        
         # if size is already computed update csize
         # otherwise the subclass shall call update_csize explicitly.
         if self.size is not NotImplemented:
             self.update_csize()
 
-        self._overrides = {}
-
+    @property
+    def cache(self):
```
rainwoodman: use_cache?

nbodykit/source/particle/uniform.py
```diff
@@ -3,38 +3,53 @@
 
 import numpy
 
-class UniformParticles(ParticleSource):
+class RandomParticles(ParticleSource):
```
rainwoodman: Difference between RandomParticles and UniformParticles?

nbodykit/tests/test_bianchi.py
```diff
@@ -0,0 +1,57 @@
...
+    NBAR = 1e-4
+    
+    CurrentMPIComm.set(comm)
+    cosmo = cosmology.Planck15
+    
+    data = Source.RandomParticles(NDATA, seed=42)
+    randoms = Source.RandomParticles(NDATA*10, seed=84)
+    
+    # add the random columns
+    for s in [data, randoms]:
```
rainwoodman: I see why RandomParticles -- very unfortunate the randoms depend on number ranks -- is there a way to wrap rng to avoid this?

nbodykit/base/particles.py
```diff
@@ -119,9 +148,8 @@ def attrs(self):
...
             self._attrs = {}
             return self._attrs
-
-    @staticmethod
-    def compute(*args, **kwargs):
+            
+    def compute(self, *args, **kwargs):
         """
         Our version of :func:`dask.compute` that computes
         multiple delayed dask collections at once
```
rainwoodman: I find myself frequently forget using this function! Shall we change the type of __getitem__ to ensure the .compute call fails?

nbodykit/base/particles.py
nickhand: ah yes

nbodykit/base/particles.py
nickhand: perhaps. or maybe just a warning? It is sometimes nice when the data is small and you have a single dask array just to call .compute(). Also, shall we change the behavior so compute() does not return tuples of length one (this seems pretty annoying)?

nbodykit/tests/test_bianchi.py
nickhand: We could certainly wrap this and hide a collective GatherArray/ScatterArray call, so that rank 0 is always doing the random number generation. That might be a nice compromise. Shall I try that?

nbodykit/base/particles.py
rainwoodman: 1. We don't know when not to issue a warning. Let's keep it as it is then.
2. Sure. tuples of length one is inconsistent with *args.

nbodykit/tests/test_bianchi.py
rainwoodman: What about generate random seeds for every 100,000 particles, then each rank generates only the overlapped chunks, and keeps only overlapping particles?

nickhand: Yes, this will work I think. Multiple ranks will generate the same samples in the overlap regions, but that waste is probably unavoidable. I'll code this up

nickhand:
okay I think this is ready to merge. The error seems to be related to too much verbosity in the logging. Shall we merge, @rainwoodman ?

rainwoodman:
Yes.

PR #293, opened 2017-01-13
WIP: implementing HOD Sources

nickhand:
Works pretty well so far. A few things that need work:

- [x] Standardization of units, especially Velocity; HOD needs km/s and Mpc/h it seems. Perhaps we could do something like Velocity [km/s] and VelocityOffset [Mpc/h]
- [x] HaloFinder only returns objects on root rank; ideally HOD would like the halos Scattered evenly
- [x] interface between HaloFinder and HaloCatalog; or maybe a different name? 
- [x] also need to implement RSD


nickhand:
The FOF halos are now in the same units as the input catalog, which will break previous results....

nickhand:
I think this is just about ready to merge....@rainwoodman, thoughts?

PR #300, opened 2017-01-27
Add a preview method for obtaining a low resolution mesh per rank.

rainwoodman:


rainwoodman:
@nickhand I am changing the normalization of LinearMesh to 1 + \delta to match it with ParticleMeshSource -- objections?

nickhand:
sounds good to me!

PR #308, opened 2017-02-03
new nightly build and activate on NERSC

nickhand:
Each night we build:
* source and dependencies of latest pypi package on 0.1.x, 0.2.x versions
* source and dependencies from HEAD of master (marked as `dev`) 
* currently no stable pypi packages with 0.2.x, so we are building 0.1.8 and HEAD

New activate.sh API:
* Activate.sh can take version number, i.e., `0.1`, `0.2`, etc or `dev` to load the HEAD of master
* Right now the allowed arguments would be `dev` or `0.1`
* No argument prints a warning and loads version `0.1`

@rainwoodman , does this work for you?


nersc/activate.sh
```diff
@@ -52,10 +86,10 @@ function srun-nbkit {
         np="-n $1"
         shift
     fi
-    if [[ "$version" != "0.2" ]]; then
+    if [[ "$version" == "0.1" ]]; then
         srun $np python-mpi /dev/shm/local/bin/nbkit.py $*
     else
-        echo "calling 'nbkit.py' is deprecated!"
+        echo "error: calling 'nbkit.py' is deprecated!"
```
rainwoodman: Maybe add a pointer to the alternative with the error message?

nickhand: good idea. added some more info. merging now

PR #313, opened 2017-02-11
fixing travis to use conda install of reqs

nickhand:


nickhand:
@rainwoodman it seems cleaner to actually use the requirements files still. In travis, we can install the requirements using conda directly. Does this look okay to you?

PR #322, opened 2017-03-31
WIP: Fgpa

rainwoodman:
This replaces #316


nickhand:
Where are we on this? Does this just transform the painted mesh by the FPGA formula?

rainwoodman:
Lukas has a version that works, but I lost track of him; this doesn't have to go to the paper.

nickhand:
use https://github.com/sbird/fake_spectra; not including here

Issue #318, opened 2017-03-07
add selection in 3pt algorithm


nickhand:
forgot to add selection support here

I am becoming more convinced from talking to new users that the ability to slice sources with a boolean array directly would be very powerful  and more intuitive....thoughts @rainwoodman?

rainwoodman:
I agree. Propagating the type of source will be nasty -- maybe we don't need that. We just need a SlicedSource type where all attrs are copied?

nickhand:
closed by #357

PR #323, opened 2017-04-01
WIP: Reorganize tests

nickhand:
The idea is to establish more comprehensive tests for each module, in addition to the "lab" tests which serve more as examples.

Also, some more documentation here. 

Thoughts on this layout, @rainwoodman?


nickhand:
this also address #318 partly, making Source sliceable with boolean arrays or integer lists

nickhand:
I like the new names with `Catalog` appended. The only remaining item that seems a little unclear to be is `CatalogMesh` (the new `ParticleMesh`)....I think I actually like CatalogMesh better than ParticleMesh, but it still feels unclear to me

PR #325, opened 2017-04-04
.apply is no longer inplace. It creates a view with extended 'actions'.

### LINKED ISSUE #324: mesh.apply shall return a new Mesh object.

rainwoodman:
Closes #324; but this is an incompatible change.


rainwoodman:
The other reason is that `[]` suggests the operation is stateful, `.` suggests the operation is stateless.

@nickhand what do you think?

nickhand:
yes this seems reasonable to me

rainwoodman:
Merge or after the test rework?

nickhand:
let's merge the test rework and then this

PR #327, opened 2017-04-06
Dataset tests

nickhand:


nickhand:
@rainwoodman - what are your thoughts on including small test data files in a "data" directory in the test directories, if needed? I think as long as we keep them small, this is cleaner than any other solutions

I've included a few small power spectra result files in `nbodykit/tests/data` to test `DataSet` properly

rainwoodman:
Yes. I agree. Paths will be hard. Are we sure python always install package data to the relative location -- or does some distributions move package data to /usr/share?

Also, I don't remember if you also need to add the directory to MANIFEST file in order to properly build a source tarball.

rainwoodman:
Looks like the path is safe: (installing data to /usr/share has been disabled) http://stackoverflow.com/a/13476594. I am still not so sure about MANIFEST.in.

nickhand:
It looks like you do not need to include them in a MANIFEST.in for 2.7+ 

http://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distribute

I've confirmed running sdist includes them. Will merge now

Issue #328, opened 2017-04-12
fix attrs when saving to bigfile 


nickhand:

nickhand:
fixed by #333

PR #333, opened 2017-04-24
save source attrs as JSON string representation (#328)

nickhand:
I think this is the most straightforward short-term solution to #328. Use `nbodykit.utils.JSONEncoder` to simply write out the JSON string in the BigFile


rainwoodman:
I suggest still make best attempt to write the attrs to the header. That's more readable than the json encoding. What about only collect those that failed into json?

nickhand:
How would we detect if attributes will fail? I think BigFile succeeds in writing them to file...it is only when we read them back in BigFileCatalog that we discover that some are actually nonsense

rainwoodman:
What if bigfile fails immediately trying to assign object array to an attr?

PR #334, opened 2017-04-28
list(self) was fishy.

modichirag:
may be the reason of an OOM error reading runpb snapshots.


nickhand:
this is a reasonable change for purposes of code simplicity though

rainwoodman:
I worry the binary file may ended up giving something that doesn't iter as expected..

rainwoodman:
I'll merge this, and we see if it fixes the issue. If it indeed does there is something fishy along TPMBinaryFile interface. The OOM occured on this line.

PR #337, opened 2017-05-02
Remove cyclic reference in mesh.to_field().save.

rainwoodman:
maybe the save method shall be moved to the mesh object?


rainwoodman:
I moved the method to Mesh object from the result of `paint()` This removes the cyclic reference and seems to make more sense (serializing to a Mesh and deserializing to a Mesh).

PR #345, opened 2017-06-04
Stop using obsolete names since halotools-0.4 and removed in 0.5

rainwoodman:
halotools 0.5 removed a few classes that were deprecated since 0.4. (https://github.com/astropy/halotools/pull/709)
This commit updates the references to the newer ones.

It appears that after the commit the code works with both halotools 0.4.0 and 0.5.0


rainwoodman:
The error is due to anaconda / travis-ci. Merging. We need a new release because astropy 0.5 is out.

PR #350, opened 2017-06-09
Avoid failling collective operations in FastPM.py

rainwoodman:
Let's see if this works.


nickhand:
This does not work....it seems like it could be a bug in bigfile? One rank always makes it through the try/except, but the others never do

rainwoodman:
I thought I fixed a bug like this.. Let me take a look.

rainwoodman:
I tagged a new bigfile that may help? 0.1.34.

nickhand:
Yep that appears to fixed it. I think you need to push the tag to GitHub so it deploys to pypi. 

I'll merge this cleaner version into #349

rainwoodman:
OK. Closing this one then.

Issue #341, opened 2017-06-01
Remove or note the latest documentation does not describe the 0.2.x branch


rainwoodman:

rainwoodman:
I am not sure if I know how the documentation is built; why is read the docs not updated?

nickhand:
closed via #351

PR #351, opened 2017-06-09
adding back barebones docs for v0.2.x for setup on RTD

nickhand:
just the main page with a note that the docs have changed


nickhand:
this closes #341

PR #352, opened 2017-06-14
WIP: Finding peak properties with FOF

rainwoodman:
This PR adds PeakPosition and PeakVelocity to FOF.

The column to define the peak is passed in as peakcolumn.

I added an algorithm KDDensity for a coarse estimation of density. This can be combined with FOF to find peak positions and velocities in the catalogue. The example is in test_fof.py


rainwoodman:
Travis is all green. @nickhand how do you think of the API changes? OK to merge?

PR #354, opened 2017-06-18
Simplify casting logic in FFTPower.

rainwoodman:
FFTPower now hosts two properties, first and second, both
are asserted to be MeshSource objects.

The resampling occurs in paint. So if a CatalogMesh object has the
wrong resolution, it will first be painted to a wrong
resolution then resampled. We may be able to
tweak the paint method of a CatalogMesh object to avoid the resampling
but I don't think it is desirable; for the compensation and perhaps
other 'actions' may depend on the resolution of the intrinsic Nmesh.

### DIFF
--- a/nbodykit/algorithms/fftpower.py
+++ b/nbodykit/algorithms/fftpower.py
@@ -5,7 +5,10 @@ import logging
 from nbodykit import CurrentMPIComm
 from nbodykit.dataset import DataSet
 from nbodykit.meshtools import SlabIterator
-from pmesh.pm import ComplexField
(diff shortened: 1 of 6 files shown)


rainwoodman:
No. The names ArrayMesh and MemoryMesh may need some adjustment. They are not very clear. What about FieldMesh and ImageMesh?

nickhand:
I like ArrayMesh and FieldMesh. I like the duality of ArrayMesh and ArrayCatalog...

nickhand:
this is merged into #370

PR #361, opened 2017-07-06
adding cylindrical grouping method algorithm

nickhand:


rainwoodman:
Is the algorithm invariant with different number of ranks?

The parallel algorithm looked a bit scary. Could you add a sentence or two in the docstring / comments about the key assumption that guarantees correct result ?

nickhand:
It's not working quite yet but I think I know the issue

nickhand:
ok @rainwoodman I'll merge this if it looks good to you

PR #365, opened 2017-07-18
Delicate functions to class instead of astropy

energy-berkeley:
Class cosmological functions are added to nbodykit.
The old astropy engine is still built:
TODO: when getattr for astropy is called, need to raise a warning. Should replace with class.
TODO: declare the dependency for classylss.

### DIFF
--- a/nbodykit/cosmology/background.py
+++ b/nbodykit/cosmology/background.py
@@ -93,6 +93,8 @@ class PerturbationGrowth(object):
             )
 
     def E(self, a, order=0):
+        a = np.array(a)
+        a[a>1] = 1
         if order == 0:
             return self.efunc(1/a - 1.0)
         else:
--- a/nbodykit/cosmology/core.py
+++ b/nbodykit/cosmology/core.py
@@ -339,6 +339,24 @@ class Cosmology(dict):
         """
         return self.clbackground.hubble_function(z)
 
+    def hubble_function_prime(self, z):
+        """
+        Function giving :func:`hubble_function prime with respect
+        to the scale factor ``a``
+
+        Parameters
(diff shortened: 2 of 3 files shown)


nickhand:
Thinking a bit more about this, I think we need to sort out a few more details. It seems to me we should make classylss just the binding of CLASS -- that allows it to be incorporated into the most packages I think. And our binding has several advantages over the default one included in CLASS, right? 

We could then move some of the power spectrum/correlation function classes to nbodykit, which already has some of that functionality. The only additional functionality in classylss right now is Zeldovich power + correlation functions, which both require FFTs via FFTLog.  Perhaps we could code that up in Cython, or use Yin's Python package ``mcfit`` to do that purely in Python (or just remove that functionality?)

@rainwoodman @energy-berkeley

rainwoodman:
Yes I think our binding preserves the good bits of the C class design better than the class one. I think we can also largely bypass the class parameter parser and directly modify the variables in the config to really push the envelope. Not sure if we want to go that far yet.

I would up vote for using Yin's mcfit for the ZA part. Yin is coming back after summer and we can probably persuade him to do this?

A severe problem with class is that it doesn't allow going beyond z=0, but for those DC mode simulations we will need a curved universe and beyond z=0. I think it is an easy patch that we can add in the binding -- this is the only blocker that I am concerned. 

What I also dislike is that in class everything is interpolated even things like hubble and hubble_prime that we know the analytic formula. This introduces some <0.1% error when we drive a ODE solver with it.

A decision still has to be made how much of astropy cosmology do we want to preserve.

rainwoodman:
This is handled in #379

PR #368, opened 2017-08-01
Add a value to to_mesh.

rainwoodman:
There are two types of weights.

The original 'weight' stands for the effective number of particles.
The second weight, 'Value' stands for the field that we are averaging.

This is crucial for dealing with multiple particle species (and masses).
We revealed the issue when Biwei was looking at hydro simulations with nbodykit.

### DIFF
--- a/nbodykit/base/catalogmesh.py
+++ b/nbodykit/base/catalogmesh.py
@@ -155,6 +155,7 @@ class CatalogMeshSource(MeshSource, CatalogSource):
                 lay = pm.decompose(position, smoothing=1.0 * paintbrush.support)
                 p = lay.exchange(position)
                 w = lay.exchange(weight)
+                v = lay.exchange(value)
 
                 H = pm.BoxSize / pm.Nmesh
 
--- a/nbodykit/base/catalogmesh.py
+++ b/nbodykit/base/catalogmesh.py
@@ -16,13 +16,14 @@ class CatalogMeshSource(MeshSource, CatalogSource):
         return "(%s as CatalogeMeshSource)" % repr(self.source)
 
     # intended to be used by CatalogSource internally
-    def __init__(self, source, BoxSize, Nmesh, dtype, weight, selection, position='Position'):
+    def __init__(self, source, BoxSize, Nmesh, dtype, weight, value, selection, position='Position'):
         # ensure self.comm is set, though usually already set by the child.
         self.comm = source.comm
         self.source = source
         self.position  = position
         self.selection = selection
         self.weight    = weight
+        self.value      = value
 
         self.attrs.update(source.attrs)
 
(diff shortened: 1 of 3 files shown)


rainwoodman:
I am a bit confused about FKP. The internal state (whether as random or as data) felt a bit unnatural. I got the code running but I am not sure what it means to assign a 'field' to fkp.

It could be one of the weights in FKP actually is the 'value' weight. Do you see it this way?

nickhand:
Sorry, what is the use case here? I am still a bit confused why we need two?

rainwoodman:
If the particles are of different mass, then we need to modify the first type of weight. Velocity is the second type of weight.

For example if the simulation has hydro particles with mass of 1/8th of the dm particles and want to compute the full power spectrum of both particle types. 

The current code will give a wrong estimation of density if we simply pass weight=mass:

```
n = n_dm * mass1 + n_hydro * mass2

nbar = N_dm + N_hydro

1 + delta = n / nbar # wrong!
```
The only way to work around this is to renormalize  the masses to weight such that sum mass == N_dm + N_hydro.

After the PR, n becomes rho:

```
n = n_dm * mass1 + n_hydro * mass2

nbar = N_dm * mass1 + n_hdyro * mass2

1 + delta = n / nbar
```

For velocity weighting, 
```
wv = n_dm * mass1 * v1 + n_hydro * mass2 * v2

nbar = N_dm * mass1  + N_hydro + mass2

1 + delta = wv / nbar 
```
is the weighted average of velocity.

nbodykit/base/catalogmesh.py
```diff
@@ -122,38 +124,46 @@ def to_real_field(self):
...
 
+            if value is None:
+                value = numpy.ones(len(position))
+
             if selection is not None:
                 position = position[selection]
                 weight   = weight[selection]
+                value     = value[selection]
+
+            Nlocal += value.sum()
```
nickhand: Shouldn't this be the sum of the Weight column? If Weight is like Mass? Or am I getting that wrong?

nbodykit/source/catalog/fkp.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, source, BoxSize, Nmesh, dtype, selection, weight, comp_weight
         self.fkp_weight  = fkp_weight
         self.nbar        = nbar
 
-        CatalogMeshSource.__init__(self, source, BoxSize, Nmesh, dtype, weight, selection)
+        CatalogMeshSource.__init__(self, source, BoxSize, Nmesh, dtype, weight, 'Value', selection)
```
nickhand: I wouldn't change FKP at all...it is sort of already handling Weight and Value, which are basically FKPWeight and CompletenessWeight. It follows the nomenclature used in the field for this so I think it's best to leave it as is. We'll document this in the docs.

nickhand:
@rainwoodman  I am going to merge this locally into #370 and add some docs about

nickhand:
Also, @rainwoodman , in the hydro example, how are you painting columns to the same field? We could allow specifying multiple columns for "position", "weight", etc in to_mesh() and paint all columns to the same mesh...

rainwoodman:
yes it should be sum of weight. i wwnt backward and forward with the names
and confused myself.

rainwoodman:
no. we have a catalog with particles of different masses. but we are not
touching hydeo only columns or reading them aa separate catelogs. I think
we can deal with this in combinecatalog, add an argument for columns to
concatenate.

Yu

nickhand:
this is merged into #370

PR #370, opened 2017-08-02
first large import of new docs

nickhand:
The amount of changes is getting large so better to merge the new docs in to master periodically. We can keep the ``docs`` branch and continue to merge things in...


nickhand:
@rainwoodman, take a look at 5f73211. It has a solution for multiple species that could work pretty well....and the same code is re-used for FKP ('data' and 'randoms'). The commit also lets you slice catalog sources using a list of column names. So you could envision doing:

```python
fullcatalog = ...
dm_cols = [...]
hydro_cols = [...]
dm = fullcatalog[dm_cols] # dm holds dm_cols + default cols
hydro = fullcatalog[hydro_cols]

species = MultipleSpeciesCatalog(['dm', 'hydro'], dm, hydro)
mesh = species.to_mesh(Nmesh=...)
real = mesh.paint() # 1+delta for n = n_dm + n_hydro
```

PR #379, opened 2017-08-29
WIP: new cosmology module using classylss + mcfit

nickhand:


nbodykit/cosmology/background.py
```diff
@@ -137,7 +137,8 @@ def _solve(self):
             raise ValueError('Neither matter or radiation dominated initial condition. Om(a_i) = %g.' % Om
                            + 'This shall not happen for a reasonable cosmology.')
 
-        y = odeint(self.ode, y0, self.lna, atol=0)
+        # solve with critical point at a=1.0, lna=0.
+        y = odeint(self.ode, y0, self.lna, tcrit=[0.], atol=0)
```
rainwoodman: How does this work exactly? later on we use self.lna so tcrit is not really computed in the results? Then the normalization at z=0 will be wrong.

nickhand: I was going to ping you about this, since I am unfamiliar with this code... ``tcrit`` specifies "critical points" where the integration "takes special care" -- I found that without it, the integration steps to negative z (which we could get around using a_max>1), but that issue is avoided by setting lna=0 as a critical point.

requirements.txt
```diff
@@ -14,3 +14,5 @@ cachey
 sympy
 numexpr
 Corrfunc
+mcfit
```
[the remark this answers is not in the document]
nickhand: they are in "requirements-extra.txt"

nbodykit/cosmology/background.py
```diff
@@ -96,17 +96,16 @@ def gf(self, a):
...
 
     def Hfac(self, a):
         return -2. - self.E(a, order=1) / self.E(a)
 
     def Om(self, a):
-        #assert abs(self.Om0 * a ** -3 / self.E(a) **2 - self.cosmo.Om(1.0 / a - 1)) < 1e-6
-        return self.cosmo.Om(1.0 / a - 1)
-        #return self.Om0 * a ** -3 / self.E(a) **2
+        z = 1./a-1
+        return self.cosmo.Omega_b(z) + self.cosmo.Omega_cdm(z) # non-relativistic
```
nickhand: Is using only Omega_b + Omega_cdm right here, @rainwoodman ? It yields better agreement in the tests, due to the different Om() definitions, I think

nbodykit/cosmology/background.py
rainwoodman: I see. The special care probably means not to jump beyond it during the integration in a poorly defined manner.

nbodykit/cosmology/background.py
rainwoodman: I think this is a good idea. The purpose of the ode solver is to describe the linear behavor of a traditional nbody simulation that that only includes perturbations of cdm and baryons.

rainwoodman:
It's all green. Shall we merge?

nickhand:
Yes, I'll merge and tag a new release

Issue #382, opened 2017-08-30
Clone is buggy, and convoluted logic in argument handling in classylss cosmology


rainwoodman:

rainwoodman:
- The temporary cosmo object created during clone is quite hard to grab. I think we are really just looking at 'equivalent classes' of arguments, and drop the ones in the old args if new ones are given.
- the semantics of cosmo.args was not clear. I think we can define it as the set of arguments to sufficiently decide a 'Cosmology' object. Then there is no need for an kwargs argument. 

I am working on a PR for this.

nickhand:
Yes I was working towards those goals I think. The args are supposed to hold named arguments and kwargs hold everything else. I think the confusion might be a remnant of when I had the "extra" dict as a keyword

nickhand:
closed by #386

PR #383, opened 2017-08-31
Cosmology clone

rainwoodman:


rainwoodman:
This changes some of the API -- as cosmology instance is made immutable (since usually makes the logic of serialization etc simpler)

I find it a bit strange LinearPower doesn't `match` sigma8 of the cosmology, but HaloFit does. I am not sure if this is intended.

rainwoodman:
This will close #382

rainwoodman:
The only down side is Cosmology(**dict(c)) won't work, but it shall be Cosmology.from_dict(dict(c)) -- given that `__init__` doesn't strongly suggest a dict as input it is probably not too bad of a compromise for avoiding keeping both args and pars.

nbodykit/cosmology/cosmology.py
```diff
@@ -35,8 +37,10 @@ class Cosmology(object):
       inferred by the curvature condition.
     * Non-cosmological constant dark energy can be used by specifying the
       ``w0_fld``, ``wa_fld``, and/or ``Omega_fld`` values.
-    * The ``sigma8`` attribute can be set to the desired value, and the internal
-      value of ``A_s`` will be automatically adjusted.
+    * To pass in CLASS parameters that are not valid Python argument names, use
+      the dictionary /keyward arguments trick, e.g.
+      `Cosmology(..., **{'temperature contributions': 'y'})`
+    * ``Cosmology(**dict(c)) is not supposed to work; use ``Cosmology.from_dict(dict(c))``.
```
nickhand: Can you use double ticks for the Cosmology(..., **{'temperature contributions': 'y'}) line? And close the ticks on Cosmology(**dict(c)) so the docs look okay?

nbodykit/cosmology/cosmology.py
```diff
@@ -10,6 +10,8 @@ class Cosmology(object):
     A cosmology calculator based on the CLASS binding in :mod:`classylss`.
 
     It is a collection of all method provided by the CLASS interfaces.
+    The object is immutable. To obtain an instance with a new set of parameters
+    use :method:`clone` or :method:`match`.
```
nickhand: Should we add an explicit __setattr__ that raises an Exception? I think you can currently set attributes, although it doesn't do anything.

nickhand:
For LinearPower, sigma8 just re-scales the amplitude, so we can don't need to keep re-computing the ClassEngine when sigma8 changes. And we also want the sigma8 attribute to agree with the sigma_r(8) function at z=0.

nickhand:
We should also fix k=0 in linear power so it doesn't raise a true_divide warning

nickhand:
merging this into a bccp:cosmology-clone branch

nickhand:
see #386

PR #386, opened 2017-08-31
Continuation of Cosmology clone (#383)

nickhand:


nickhand:
This is all clear now and I think includes all of the desired fixes from the original PR @rainwoodman

rainwoodman:
What about adding an `match(Omega_cb=?)`

nickhand:
What is ``Omega_cb``?

rainwoodman:
Like Omega_m in Astropy; `match` can adjust Omega_cdm, Omega_b,  and Omega_lambda to match a given Omega_cb if Omega_k was zero?

nickhand:
Ah, I see what you mean...do you like ``Omega_nr``, as in non-relativistic at all times, any better?

nickhand:
Is it worth it to explore matching Onu? The result of the to_astropy() function has some discrepancies due to differences in the neutrino definitions. I had a quick look and it seemed like it would be difficult to match properly

rainwoodman:
Isn't Omega_nr just Omega_m ?

The time dependency of Onu is different, so it is impossible to match them..

nickhand:
I guess I think of Omega_nr as only CDM + baryons,  the species that are always cold, whereas Omega_m has the non-CDM too...

rainwoodman:
A portion of the ncdm distribution function is nr too, so...

nickhand:
Yep I agree. I was just trying to think of a better name. This looks good though

nickhand:
It might be worth it to include deprecated support for the past Cosmology() signature. I am realizing that past Cosmology() inits are broken now

nickhand:
This should be all green now once we tag a new classylss version. @rainwoodman let me know what you think of the deprecated __init__ handling. I think the only downside is we lose the function signature for Python 2 users

rainwoodman:
It feels like AstropyCompat shall be moved to nbodykit too.

rainwoodman:
Shall we go ahead with a new classylss tag?

nickhand:
Yes, I'll tag a new classylss version now. Let's leave AstroCompat where it is for now

nickhand:
Shall we merge this @rainwoodman? I have a couple of other PRs to put in and then we can tag a new release I think.

rainwoodman:
Sure!

PR #388, opened 2017-09-01
Param with zeros

rainwoodman:
I am not sure how desirable this is. I do find it a bit confusing that Omega_cdm in __init__ becomes Omega0_cdm in the attribute.


nickhand:
Yes I agree it is annoying that the CLASS input module prefers without the 0. This seems like a reasonable solution...

rainwoodman:
I'll merge this in then.

PR #389, opened 2017-09-05
Fixing N_ur to match class paramters.

energy-berkeley:
As in title, using the correspondence of no. of massive neutrino species and N_ur in class

### DIFF
--- a/nbodykit/cosmology/cosmology.py
+++ b/nbodykit/cosmology/cosmology.py
@@ -94,8 +94,8 @@ class Cosmology(object):
             T_cmb=2.7255,
             Omega_b=0.022032/0.67556**2,
             Omega_cdm=0.12038/0.67556**2,
-            N_ur=None,
-            m_ncdm=[0.06],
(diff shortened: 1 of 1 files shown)


rainwoodman:
The reason `N_ur` is None is because if not give, we find the correct number from the table  in `compile_args`.

https://github.com/bccp/nbodykit/pull/389/commits/5eb1a2d77d74c86efef7c215ef8d9b5015d1d7fe#diff-802eaea1e0b6003b0dc7ac4a26b7b0a6L477

It is done this way to avoid introducing a dependency between `N_ur` and `m_nu`. If we do it your way then whenever `m_nu` is set to a different length, a different value of `N_ur` has to be specified -- one more rule to follow and one more place to make mistake.

energy-berkeley:
I see, so, if I were to disable neutrinos, I would need m_ncdm = None, because setting m_ncdm = [0] returns error.

rainwoodman:
you meant disableing massive neutinos. yes

Issue #377, opened 2017-08-25
CatalogCopy better named to CatalogRef?


rainwoodman:

nickhand:
Yes, I think I like this name better...

nickhand:
The problem is that it's not always a subset, like when we are concatenating multiple source together. It's really just a Catalog where the columns are passed in as kwargs

rainwoodman:
Maybe we don't need a new class? It's just a special method of Catalog?

rainwoodman:
I am also a bit unsure the relation between this and an ArrayCatalog, which takes an dictionary as an input but does a bunch more (seemingly unnecessary) work.

nickhand:
I think the difference is that the size is known beforehand for CatalogCopy/Ref but not for ArrayCatalog . But we could remove CatalogCopy in favor of a class method of CatalogSource -- it makes more sense to have "CatalogSource" then showing up in all of the log statements

nickhand:
I guess we'll need to get rid of the abstract property tag on ``size`` for this to work?

rainwoodman:
Yes - I did that in #390.

nickhand:
closed via #390

PR #390, opened 2017-09-08
rename catalogcopy -> catalogsubset

rainwoodman:
because no copy is ever made.


rainwoodman:
per #377, I'll squash the class name, and use a class method instead.

nickhand:
this looks good to me!

nickhand:
I'm playing around the FKP / MultipleSpecies interface so I'll merge this

Issue #391, opened 2017-09-08
optimize selection process?


nickhand:
I ran into an annoying use case where an input catalog included invalid data (in this case, negative redshifts). I sliced the object to include only valid data, but only after doing my data cleaning. In this case, I added "Position" (which cannot be computed for z<0). When computing Position, this fails because the selection is the last step performed (the dask engine does all of the operations on the whole catalog first).  You can say this is a user error (I should have sliced first) but I would argue it's likely a common mistake and one that would be nice to optimize away.  

In theory, this can be optimized away for potentially significant gains. The size of the catalog is fixed after some point in the dask graph, so we can insert the selection tasks into the dask graph after the size is finalized. 

I am working on a potential implementation of this -- it delves into the guts of dask a bit so we'll need to see if it is worth it.

rainwoodman:
I still don't quite understand this. could you post a code segment?

nickhand:
Currently this fails:

```python
s = RandomCatalog(1000, seed=42)
s['z']   = s.rng.normal(loc=0, scale=0.2, size=s.size) # contains z < 0
s['ra']  = s.rng.uniform(low=110, high=260, size=s.size)
s['dec'] = s.rng.uniform(low=-3.6, high=60., size=s.size)

# add Position
s['Position'] = transform.SkyToCartesion(s['ra'], s['dec'], s['z'], cosmo=cosmology.Planck15)

# select valid range
subset = s[s['z'] > 0]

# fails with CLASS error for z<0 
pos = subset['Position'].compute()
```

rainwoodman:
I see -- I was confused with the selection column. The selection is performed after compute. I'll take a better look at 392 tomorrow.

nickhand:
closed via #392 included as part of #393

PR #392, opened 2017-09-09
Dask optimizations 

nickhand:
This adds two main features:
1. A ``set_options`` context manager (as in dask) to specify default dask chunk size and cache sizes. I've also increased the default chunk size to 5e6 -- dask docs recommend chunk sizes holding between 10-100 MB to avoid overhead. I definitely see overhead delays when reading large randoms catalogs on my laptop with the old chunk size (1e5)
2. #391 :  An optimized selection procedure for selecting subsets of CatalogSource objects -- selection tasks are inserted into the task graph at the earliest point once the CatalogSource size is fixed. So operations on columns will be performed on the selected subset automatically.

### DIFF
--- a/nbodykit/__init__.py
+++ b/nbodykit/__init__.py
@@ -6,6 +6,33 @@ from mpi4py import MPI
 import dask
 dask.set_options(get=dask.get)
 
+_globals = {}
+_globals['dask_cache_size'] = 1e9
+_globals['dask_chunk_size'] = int(5e6)
+
+class set_options(object):
+    """
+    Set global configuration options.
+
+    Parameters
+    ----------
+    dask_chunk_size : int
+        the number of elements for the default chunk size for dask arrays;
+        chunks should usually hold between 10 MB and 100 MB
+    dask_cache_size : float
+        the size of the internal dask cache in bytes; default is 1e9
+    """
+    def __init__(self, **kwargs):
+        self.old = _globals.copy()
+        _globals.update(kwargs)
+
+    def __enter__(self):
+        return
+
+    def __exit__(self, type, value, traceback):
+        _globals.clear()
+        _globals.update(self.old)
+
 class CurrentMPIComm(object):
     """
     A class to faciliate getting and setting the current MPI communicator.
--- a/nbodykit/base/catalog.py
+++ b/nbodykit/base/catalog.py
@@ -1,13 +1,62 @@
 from six import add_metaclass, string_types
 from nbodykit.transform import ConstantArray
+from nbodykit import _globals
 
 import abc
 import numpy
 import logging
 import warnings
+from collections import defaultdict
 import dask.array as da
-from nbodykit import _globals
 
+def get_slice_size(start, stop, step):
+    """
+    Utility function to return the size of an array slice
+    """
+    if step is None: step = 1
+    if start is None: start = 0
+    N, remainder = divmod(stop-start, step)
+    if remainder: N += 1
+    return N
+
+def isgetitem(task):
+    """
+    Return True if the task is a getitem slice
+    """
+    from dask.array.optimization import GETTERS
+
+    if isinstance(task, tuple) and task[0] in GETTERS:
+        sl = task[2][0] # first dimension slice
+        if sl is not None:
+            return True
+    return False
+
+def find_consecutive_gets(t, out, dsk, dependents):
+    """
+    Fill the defaultdict ``out`` with consecutive getitem tasks that have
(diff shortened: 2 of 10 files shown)


nickhand:
Any thoughts on this @rainwoodman ?

nbodykit/base/catalog.py
```diff
@@ -1,14 +1,195 @@
...
+    if not len(dsk2):
+        raise ValueError("cannot perform optimized selection")
+
+    # update the original graph and make new Array
+    dsk.update(dsk2)
+    chunks = tuple(blockdims) + arr.chunks[ndim:]
+    return da.Array(dsk, name, chunks, dtype=arr.dtype)
+
 
 class ColumnAccessor(da.Array):
```
rainwoodman: This requires dask > 0.14.2. I just found yesterday.

nbodykit/base/catalog.py
```diff
@@ -35,6 +216,36 @@ def __new__(cls, catalog, daskarray):
...
+        d = None
+
+        # try to optimize the selection
+        sel = key[0] if isinstance(key, tuple) else key
+        if isinstance(sel, (list, numpy.ndarray, da.Array)):
+            if len(sel) == self.catalog.size:
+                if isinstance(sel, da.Array):
+                    sel = self.catalog.compute(sel)
+                try:
+                    d = optimized_selection(self, sel)
```
rainwoodman: This try except looks strange -- could you catch a DaskGraphOptimizationFailure instead?

nbodykit/__init__.py
```diff
@@ -6,6 +6,33 @@
 import dask
 dask.set_options(get=dask.get)
 
+_globals = {}
```
rainwoodman: _global_options?

nbodykit/base/tests/test_catalog.py
```diff
@@ -1,12 +1,59 @@
...
+
+    CurrentMPIComm.set(comm)
+
+    # compute with smaller chunk size to test chunking
+    with set_options(dask_chunk_size=100):
+
+        s = UniformCatalog(1000, 1.0, seed=42)
+
+        # slice once
+        subset1 = s[:20]
```
rainwoodman: Is this :20 per rank or :20 of the entire dataset?

rainwoodman: Maybe we shall define gslice(a, b, c) that represents global selection cross ranks? It would be useful in abundance matching a sorted catalog.

nbodykit/tests/test_lab.py
```diff
@@ -53,3 +53,20 @@ def filter(k, v):
 
     source.save(output="./test_paint-real-%d.bigfile" % comm.size, mode='real')
     source.save(output="./test_paint-complex-%d.bigfile" % comm.size, mode='complex')
+
+@MPITest([1, 4])
+def test_set_options(comm):
+
+    CurrentMPIComm.set(comm)
+
+    with set_options(dask_cache_size=5e9, dask_chunk_size=75):
```
rainwoodman: Can I use set_options not as a context manager? The name doesn't suggest it is a context manager at all.

rainwoodman: If it cannot be used without as a context manager, then it can be renamed to `option_context`.
```
with nbkit.option_context(....) as ctx:
    ...
    ...
```

nbodykit/base/catalog.py
nickhand: I'll pin the version in requirements.txt

nbodykit/base/catalog.py
nickhand: yeah that is a clearer solution

nbodykit/__init__.py
nickhand: yes, I'll change it

nbodykit/base/tests/test_catalog.py
nickhand: Yes, currently the slicing is all local. A global slice would be nice. I think I have some code I started on for that already. I'll open up a new PR for it

nbodykit/tests/test_lab.py
nickhand: Yes it will set the globals if it is not called as a context manager. It largely follows the dask ``set_options`` class.

rainwoodman: I guess it probably came from numpy's formatting options -- though I am still a bit reserved on their choice of name -- a context named like an action. But looks like we are stuck with such names.

PR #393, opened 2017-09-14
WIP: Implement CatalogMesh as a view of CatalogSource

nickhand:
The old CatalogMesh was unclear about who owned the columns, with unusual side-effects when adding columns directly to a mesh. This PR aims to implement CatalogMesh as a view of a CatalogSource object, similar to the numpy "view". This also implements slices of CatalogSource objects as views as well.  

Also: 
- added functionality to ConvolvedFFTPower for cross-correlations in this PR. 
- includes #392 b/c some of the new functionality depended on it


nickhand:
There's still a lot to be cleaned up here...in particular, I wasn't sure how to best init the MeshSource when initializing a new CatalogMesh from __new__. Thoughts, @rainwoodman ?

nickhand:
FYI, from 57ed564 onwards are the new commits implementing the view...

nbodykit/base/mesh.py
```diff
@@ -34,44 +34,67 @@ def __init__(self, comm, Nmesh, BoxSize, dtype):
...
-        self.attrs['BoxSize'] = self.pm.BoxSize.copy()
-        self.attrs['Nmesh'] = self.pm.Nmesh.copy()
+        if hasattr(obj, 'pm'):
+            self.pm = obj.pm
+        else:
+            initialize_pm(self)
 
         self._actions = []
-        self.base = None
+        self.actions.extend(getattr(obj, 'actions', []))
```
rainwoodman: This is heading towards the direction of having 'actions' as a part of dask diagram..

nbodykit/base/mesh.py
```diff
@@ -359,3 +371,28 @@ def save(self, output, dataset='Field', mode='real'):
                             bb.attrs[key] = json_str
                         except:
                             warnings.warn("attribute %s of type %s is unsupported and lost while saving MeshSource" % (key, type(value)))
+
+
+def initialize_pm(self):
```
rainwoodman: This is not a member of the class?

nbodykit/algorithms/convpower.py
```diff
@@ -669,21 +669,23 @@ def normalization(self, name, alpha):
 
         if name+'.norm' not in self.attrs:
 
-            # first/second meshes for "name" (data or randoms)
-            first = self.first[name]
-            second = self.second[name]
+            # the selection (same for first/second)
+            sel = self.first[name][self.first.selection]
```
rainwoodman: is self.first.selection an array like?

nbodykit/base/catalogmesh.py
```diff
@@ -38,66 +39,69 @@ class CatalogMesh(MeshSource, CatalogSource):
...
-        return self.source.hardcolumns
+    def __init__(self, *args,  **kwargs):
+        pass
 
-    def get_hardcolumn(self, col):
-        """
-        Return a hard-coded column
-        """
-        return self.source.get_hardcolumn(col)
+    def __finalize__(self, obj):
```
rainwoodman: Why not using  __init__? Another possibilty is to use `setstate`

I think if __new__ exists you have to explicitly trigger __init__ anyways.. (I may be wrong).

nickhand: ``obj`` here can be a CatalogSource or a CatalogMesh, so I'm not sure ``__init__`` would work. I guess this is sort of a ``setstate``.

``CatalogMesh.__finalize__`` will get called in different situations: 

1. when slicing a CatalogMesh to get a subset

https://github.com/bccp/nbodykit/blob/884f57ca748149fe7871138b01285d4c65fdea88/nbodykit/base/catalog.py#L434

2. when initializing a CatalogMesh from a CatalogSource

https://github.com/bccp/nbodykit/blob/884f57ca748149fe7871138b01285d4c65fdea88/nbodykit/base/catalogmesh.py#L53

nbodykit/algorithms/convpower.py
nickhand: ``sel`` is a dask array. ``self.first.selection`` is the name of the selection column, and ``self.first[name]`` is a CatalogMesh holding the data or randoms species

nbodykit/base/mesh.py
nickhand: It's not right now but it could be. But this is part of the issues I was having with MeshSource. Ideally, we should probably have some sort of a mix-in MeshSource class  to CatalogMesh that doesn't need to be initialized, but we need to work around the initializing ``pm`` and using the vector forms of ``Nmesh`` and ``BoxSize`` from the ``pm`` attribute

nbodykit/base/mesh.py
nickhand: How so? You can't have actions applied after the ``r2c`` that way?

rainwoodman: More or less like dask.delayed?

nbodykit/base/catalogmesh.py
rainwoodman: I guess what makes me confused is the function is defined by the occasion it is called, but not by its purpose -- thus it starts to feel like we are building a framework.

If I read the function, the purpose of the function is to copy Python attributes? would it work if we simply copy the `__dict__` of the object instead?

nickhand: The logic is the same as numpy's ``__array_finalize__``. In this case, copying ``__dict__`` won't work because when you are going from CatalogSource to CatalogMesh, ``self`` in ``__finalize__(self, obj)`` is a bare object, and you need to add the default CatalogMesh attributes to it.

nickhand:
Okay I re-worked some of the logic to make more sense now. ``__finalize__`` just attaches attributes from one object to another, and allows us to return slices of the same type and implement CatalogMesh as a view of CatalogSource. 

Let me know what you think @rainwoodman -- I think this is ready for merging soon

nbodykit/algorithms/tests/test_conv_power.py
```diff
@@ -201,13 +313,13 @@ def test_with_zhist(comm):
 
     # add n(z) from randoms to the FKP source
     nofz = InterpolatedUnivariateSpline(zhist.bin_centers, zhist.nbar)
-    fkp['randoms/NZ'] = nofz(randoms['z'])
-    fkp['data/NZ'] = nofz(data['z'])
+    fkp['randoms']['NZ'] = nofz(randoms['z'])
```
rainwoodman: Does this mean the slash syntax is replaced by more brakets?

rainwoodman:
Yes I think this is a very big PR and in a good direction. Do we still support the slash syntax looking up sub-catalogs in a multispecies catalog? I noticed you seem to have changed the syntax to `[][]`.

nbodykit/algorithms/tests/test_conv_power.py
nickhand: Both are supported now and give identical results -- sometimes it feels easier to slice with the species name first when the species name is a variable

nickhand:
Okay I am going to go ahead and merge this

PR #396, opened 2017-09-18
Add some notes about login nodes.

rainwoodman:
Does this look right?

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -159,6 +159,16 @@ the ``--user`` or ``--root <dir>`` options of the ``pip install`` command.
 nbodykit on NERSC
 -----------------
 
+.. notes::
+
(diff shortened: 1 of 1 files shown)


rainwoodman:
I remember clearly seeing Chirag running nbodykit from the login node with the bccp channel binary (and mpich from anaconda). It didn't work because of the GLIBC version issue, but Cray updated to GLIBC 2.22 'very recently' ( less than a month?)

And yes, never leave the notebook again -- except waiting for the jobs to finish.. jobs has to talk to notebook via files -- this is where BigFileCatalog and BigFileMesh becomes handy I assume.

nickhand:
@rainwoodman you are right, it works on the login nodes with anaconda. I've updated the notes

Issue #397, opened 2017-09-19
global slice and sort functions


nickhand:
This is relatively easy to add and will be a nice feature for users to have, especially for i.e., #304

nickhand:
closed via #406

PR #406, opened 2017-09-22
Global slice and sort (#397)

nickhand:
Adds global sort and global slice functions to CatalogSource (and CatalogMesh) objects


nickhand:
Is this what you had in mind @rainwoodman ?

nbodykit/base/catalog.py
```diff
@@ -1165,6 +1165,117 @@ def csize(self):
             return NotImplemented
         return self._csize
 
+    def gslice(self, start, stop, end=1):
+        """
+        Execute a global slice of a CatalogSource.
+
+        .. note::
+            After the global slice is performed, the data is scattered
```
rainwoodman: I wonder if it is useful to avoid the scattering with an option?

rainwoodman:
I have to find a larger chunk of time to read through this. But briefly reading this, it seems not supporting slicing a ColumnAccessor. Could it be useful to globally slice a columnaccessor?

nbodykit/base/catalog.py
nickhand: I am not sure I can think of a use case now, but it doesn't hurt to include the option, I think

nickhand:
No problem, let's push this to v0.2.8 anyways. 

I think I agree. It would be nice to call both sort() and gslice() on a ColumnAccessor, I think

nbodykit/base/catalog.py
rainwoodman: Avoiding the scattering probably means the compute can be deferred further?

rainwoodman:
The intended main use case of gslice is for abundance matching. Gslicing on a column accessor does not fit into this context.

Currently a gslice will trigger a full scan of the catalog, does it?

nbodykit/base/catalog.py
nickhand: yes I think so. I wonder if it is possible to implement this with no compute() calls?  We can't use Scatter/GatherArray for dask arrays but we can bcast the appropriate slices of them?

nickhand:
As it's written now, it computes the whole catalog and re-scatters it evenly after the slice. In theory, we could pass around the dask arrays instead?

rainwoodman:
Perhaps leave a few comments in the code to describe how this 'could be done' without actually doing it. Because in reality if we are abundance matching we probably can easily read in the entire data set in most cases (especially we already done a sorting) The IO is cut off from that point when we have the abundance matched catalog, and it's not a bad thing.

The situation may change if we are goaling to make catalogs for LSST or BOA.

nickhand:
This looks ready too so I'll merge this in for 0.2.8

Issue #398, opened 2017-09-19
angular correlation function


nickhand:
Can wrap Corrfunc for this as well

nickhand:
closed via #408

PR #408, opened 2017-09-26
PairCount upgrades (angular pair count + load balance)

### LINKED ISSUE #364: domain decomposition on (ra, dec, z) surveys leads to unbalanced loads
When you convert (ra, dec, z)  to Cartesian coordinates, the particles aren't uniformly distributed in the  resulting box. The consequence is that when you do the normal grid domain decomposition in ``SurveyDataPairCount``, a lot of ranks have zero or very few particles and the load is unbalanced and a few ranks are responsible for most of the work. 

@rainwoodman do you see a potential fix for this, or maybe we just want to remove the domain decomposition?

nickhand:
This re-factors pair counting algorithms with several upgrades:

- survey algorithms use the new pmesh load balancing (this fixes #364)
- angular pair count on (ra, dec) data added (#398)
- better error handling for Corrfunc calls. Corrfunc can raise both meaningful and non-meaningful exceptions, while printing error messages to screen. New code captures all output and raises reasonable exception
- run Corrfunc code in iterations (10 by default) so user has some measure of progress for long-running cases. Often, I have found myself and others unable to gauge timing of code on NERSC due to node time-outs without any progress logged


nickhand:
All the tests against kdcount have passed so I am going to merge this and have Kaylan starting playing around with it.

Issue #417, opened 2017-10-04
question: using FOF Class on observed (not simulated) galaxy catalog


moustakas:
I've just started using the awesome `nbodykit` tools -- thanks for writing them!

I'd like to use the `nbodykit.algorithms.fof.FOF` class to generate a group catalog from an observed sample of galaxies with just RA,Dec coordinates using a simple angular linking length.  I'm doing something like
```
from nbodykit.transform import SkyToUnitSphere
from nbodykit.source.catalog import FITSCatalog
from nbodykit.algorithms.fof import FOF

cat = FITSCatalog('mycatalog.fits')
cat['Position'] = SkyToUnitSphere(cat['RA'], cat['DEC'], degrees=True)
fof = FOF(cat, linking_length=0.05, nmin=1, absolute=True)
```
However, this crashes with the message
```
ValueError: cannot compute FOF clustering of source without 'BoxSize' in ``attrs`` dict
```
Is what I'm trying to do possible?

rainwoodman:
FOF requires you to embed the data to a periodic box -- in this case you can probably add
to your script
```
cat.attrs['BoxSize'] = 4.0
cat['Position'] += 2.0
```

Which will shift the points to the center of the box of size 4.0 -- Could you see if this works? We shall add a non-periodic option to FOF that uses automatically computed boxsizes. I'll file a PR to add this feature.

rainwoodman:
Feature is in (periodic == False) -- workaround above no longer needed. Closing. Thanks!

PR #418, opened 2017-10-04
Adding Non-periodic support to FOF.

rainwoodman:
This addresses #417, where a boxsize is absurdly required for a non-periodic dataset. I guess in other places of the code we still mostly assume periodic boundary. I wonder if this simplifies the FiberAssignment algorithm?

### DIFF
--- a/nbodykit/algorithms/fof.py
+++ b/nbodykit/algorithms/fof.py
@@ -351,19 +351,19 @@ def fof(source, linking_length, comm, periodic):
 
     np = split_size_3d(comm.size)
 
-    BoxSize = source.attrs.get('BoxSize', None)
-
-    if BoxSize is None:
-        if periodic:
+    if periodic:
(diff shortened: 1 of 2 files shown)


rainwoodman:
@moustakas Will you have some time to help us doing some QA-ing -- by reviewing the change set. This is a small PR and won't take long to look at.

moustakas:
Thanks for turning this around so quickly!  I was able to checkout and test this branch and the `FOF` algorithm indeed runs successfully (after setting `periodic=False`) and quite fast on an input catalog of >2.5 million objects.

My only question now how to convert the angular scale / separation I'm interested in into a linking_length in Cartesian coordinates, but that just means I need to review my spherical trigonometry.

rainwoodman:
Thanks! I'll merge this then -- the convertion is r = 2sin(theta/2). I believe.

PR #420, opened 2017-10-05
fixes #419 and remove optimized dask selection

### LINKED ISSUE #419: selection causing parallel code to hang
The issue appears to be an optimization added in 0.2.7 when we return self if the selection index does not change the catalog size.

Old code check only if the local size is unchanged, but this should be the collective size...trying out a fix.

nickhand:
This fixes two issues arising in v0.2.7 on NERSC: 

- #419: only slices that do not change the **collective** size, not local size do not need to be evaluated
- dask optimized selection is buggy and can fail in a very bad way, at the time of a compute() call, breaking basically the whole code. After some more thought, I think it's best to not fiddle with the dask internals as much as possible, given that the internal API can change at any time. Also, it's a hard piece of code to debug and maintain, so it seems best to remove for both the future usability of the code and future maintenance

Thoughts, @rainwoodman?


nickhand:
I've confirmed this fixes my recent hanging issues on NERSC, and we should push a v0.2.8 once @rainwoodman is on board.

rainwoodman:
I always prefer writing 20% of code to cover the 80% case -- so yes go ahead and remove the dask internal stuff!

PR #425, opened 2017-10-23
I suggest not using latex

rainwoodman:
It is not always properly installed on the target system. For example jupyter-dev.nersc.gov has a crippled latex and causes the notebook to fail to run. Does the font look very bad without using Tex?


nickhand:
Good catch, I do think it looks better, but mostly it was there because that's the default on my own machine

Issue #422, opened 2017-10-08
astropy methods for cosmology class


fbeutler:
I am trying to calculate the Hubble parameter with the cosmology class, but it does not seem to exist? The angular diameter distance works file

    print("cosmo.H(0.5) = ", cosmo.H(0.5))
  File "/Users/xflorian/anaconda/envs/nbodykit-env/lib/python3.6/site-packages/nbodykit/cosmology/cosmology.py", line 221, in __getattr__
    raise AttributeError("Attribute `%s` not found in any of the delegate objects" % name)
AttributeError: Attribute `H` not found in any of the delegate objects

rainwoodman:
Most of the astropy methods are removed as we change to CLASS set of function names:
```
In [3]: from nbodykit.cosmology import Planck15

In [4]: Planck15.hubble_function?
Docstring:
Background.hubble_function(self, z)

The Hubble function in CLASS units. Users should use :func:`efunc`
instead.
Type:      builtin_function_or_method
```

Tab-completion shows these names but currently they are not in the documentation of classylss or nbodykit, unfortunately.

rainwoodman:
As, as noted in the docstring, you probably want to use Planck15.efunc and multiply with Planck15.H0 (in Mpc/h units) instead.

rainwoodman:
Temporary documentation of the available methods is listed at http://rainwoodman.github.io/classylss.

fbeutler:
thanks Yu

rainwoodman:
The documentation is added. Check it out! http://nbodykit.readthedocs.io/en/latest/api/api.html#cosmology-nbodykit-cosmology

PR #426, opened 2017-10-23
update docs workflow + proper Cosmology docs

nickhand:
A few updates:

- api docs are generated with the builtin auto summary feature from a dynamically generated modules.rst file
- there is now a proper list of methods/attributes for Cosmology(), linking to classylss -- this is accomplished with a custom sphinx template passed to auto-summary
- This solves the issue of RTD timeouts (15 mins max) by using Jupyter notebooks instead of the ipython directive, which was taking up too much time when building the docs. Notebooks support RST raw cells, so we can do any modules that need code examples in notebooks instead
- fixes various deprecated syntaxes in notebooks


nickhand:
The general workflow that I am using is doing: 

``` bash
make notebooks
make html
```
which executes all *ipynb files, and then builds the docs. Then, on RTD, the ipynb files are not re-executed, so we shouldn't bump up against the timeout limit

rainwoodman:
Looks like magic, but as long as it works.

I slightly liked the .rst files better than notebooks, because they have better looking diff-sets.

nickhand:
Yeah the notebook solution is not ideal (although it's only the docs with the IPython directive that get replaced), but I think it's the cleanest solution at the moment. I imagine issues like this are why projects like yt have a dedicated machine to build the docs, custom domain, etc

rainwoodman:
screenshot from 2017-10-23 14-07-33

Here is an issue -- get_pk occured twice.. usually the one from primordial is shadowed by the one from Spectra. I think we shall rename get_pk in primordial to get_pk_primordial to resolve this conflict. Is there a way to detect these conflicts in docs?

nickhand:
ah yes, I noticed that and meant to fix that in the docs. I'll do that now

rainwoodman:
Looks like it is less code if you keep the duplicates around, but issue a warning instead? The 'dir' method of a cosmology object does return every member it can find -- will it be useful to simplify the autosummary code? would it be useful if we implement __dir__ on the class instead?

nickhand:
I think we should avoid leaving duplicates in the docs list, since that could be confusing for the user. 

For the auto summary list, I'm searching in the same manner as the Cosmology's __getattr__ (not __dir__), so now it always links to the proper attribute that the user will get. This is also an issue for sigma8, where we overload the Spectra attribute in the Cosmology() class.

nickhand:
FYI, I've built the docs with this update here:

http://nbodykit.readthedocs.io/en/cosmo-docs/index.html

nickhand:
this addresses and closes #422

PR #427, opened 2017-10-23
Add a rank column

rainwoodman:
this is useful for abundance matching.

### DIFF
--- a/nbodykit/base/catalog.py
+++ b/nbodykit/base/catalog.py
@@ -1022,13 +1022,16 @@ class CatalogSource(CatalogSourceBase):
         return ConstantArray(1.0, self.size, chunks=_global_options['dask_chunk_size'])
 
     @column
-    def Rank(self):
+    def Index(self):
         """
-        The column giving the rank of each particle in the list.
-
+        The column giving the index rank of each particle in the list. It is
+        an integer from 0 to ``self.csize``
         """
         offset = sum(self.comm.allgather(self.size)[:self.comm.rank])
-        return da.arange(offset, offset + self.size, chunks=_global_options['dask_chunk_size'])
+        # do not use u8, because many numpy casting rules case u8 to f8 automatically.
+        # it is ridiculous.
+        return da.arange(offset, offset + self.size, dtype='i8',
+               chunks=_global_options['dask_chunk_size'])
 
     @column
     def Value(self):
--- a/nbodykit/base/catalog.py
+++ b/nbodykit/base/catalog.py
@@ -1021,11 +1021,16 @@ class CatalogSource(CatalogSourceBase):
         """
         return ConstantArray(1.0, self.size, chunks=_global_options['dask_chunk_size'])
 
-    @column
+    @property
     def Index(self):
         """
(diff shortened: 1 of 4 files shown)


nbodykit/base/catalog.py
```diff
@@ -1022,6 +1022,15 @@ def Weight(self):
         return ConstantArray(1.0, self.size, chunks=_global_options['dask_chunk_size'])
 
     @column
+    def Rank(self):
+        """
+        The column giving the rank of each particle in the list.
+
+        """
+        offset = sum(self.comm.allgather(self.size)[:self.comm.rank])
+        return da.arange(offset, offset + self.size, chunks=_global_options['dask_chunk_size'])
```
nickhand: use 'u4' as the dtype?

nickhand:
I think we also use the "Rank" column here in the CGM algorithm:

https://github.com/bccp/nbodykit/blob/master/nbodykit/algorithms/cgm.py#L158

nickhand:
and we should update the docs table of default columns:

https://github.com/bccp/nbodykit/blob/master/docs/source/catalogs/overview.rst

nbodykit/base/catalog.py
rainwoodman: u8? It can be more than 4billion..

rainwoodman:
Sure. BTW an issue with these default columns is that there is no easy way to write a catalog to disk bypassing these columns if they haven't been overridden.

nickhand:
That's a good point. I guess the most sensible behavior would be not save those columns automatically and restore them when loading automatically. We would need to tag them somehow as defaults I guess..

rainwoodman:
We currently have this 'hardcolumn' thing. But I keep forgetting what it really means.

nickhand:
I believe the “hardcolumn” refers to any column that uses a method of the
class to return the column, so these default columns are “hardcolumns” but
not all hardcolumns are defaults

rainwoodman:
I have been a bit wary about this. Looks like there are two ways of dealing with hardcolumns, either via a decorator or via a two function protocol. Do you know which subclass uses the protocol method and why?

nickhand:
what do you mean by two function protocol? I think only the decorator is being used?

rainwoodman:
There is also get_hardcolumns() that states it can be overridden...

nickhand:
It looks like only the array and file catalogs overload get_hardcolumn() to dynamically get any of the fields in `dtype` when the user requests... I feel like we could probably simplify this to behave like a dictionary with default values?

rainwoodman:
It currently already behave like a dict with default values. I think what we need is a tag on a column that it shall not be saved. Alternatively speaking, we want a function that returns a list of columns that shall go to disk.

rainwoodman:
The rank column is different from others. I think it shouldn't be a column at all! Because after selection the rank changes!

nickhand:
that is an excellent point

rainwoodman:
Merge this after it is green? Chirag wants to use it.

nickhand:
sorry... yes looks good to me. I'll merge

PR #428, opened 2017-10-23
Interlaced species

rainwoodman:
See the last cells of the pdf file:

[mozilla.pdf](https://github.com/bccp/nbodykit/files/1408694/mozilla.pdf)

The mean is not properly normalized when interlacing is enabled. Looks like some of the buffer reuse is incorrect.


rainwoodman:
This PR avoids memory reuse in interlaced handling, and also lifts the multiple FFTs outside of the loop.

Issue #404, opened 2017-09-20
Add a binder shortcut.


rainwoodman:

nickhand:
I had thought of this before but they were having a bunch of issues rendering. It looks like they are upgrading

rainwoodman:
yes i was at thw and it worked pretty smooth here.

nickhand:
closed via #435

PR #435, opened 2017-10-31
Configure binder

nickhand:
Binder seems to be back to working and everything looks good. This PR adds a binder badge to the README and docs that will launch a JupyterHub instance containing all of our cookbook recipes. 

This sets up binder by doing the following:

- move the old cookbook directory to bccp/nbodykit-cookbook
- add the bccp/nbodykit-cookbook as a submodule in the docs/ directory
- make sure nbodykit-cookbook has a conda environment.yml file for installing dependencies on binder

Solves #404


nickhand:
@rainwoodman, there is definitely a downside to having the cookbook as a separate repo, but binder seems to works best with an isolated repo and I think the benefits of having binder for new users will outweigh any submodule confusion for developers. I've added some notes to the ``docs`` README about the work flow I am using...I am still a bit unfamiliar with submodules tho

rainwoodman:
submodule command names are misleading. Be sure to note that

- git submodule update : checkout the version of submodule as recorded 

- to update the ref of the submodule you have to go in the directory, git pull; then go out and git add; git commit. So it will always need a commit.

DoOne possiblity is to do more magic in readthedocs scripts to checkout a nbodykit-cookbook repo, grab the files and build. Then you don't have to deal with submodules.

nickhand:
I think I like downloading the GitHub repo at build time on RTD. I think it's a better workflow to avoid the submodule confusion and keep all changes confined to the working directory of the ``nbodykit-cookbook`` repo. Let me see if I can get that set up

nickhand:
I've got the workflow set up to download the cookbook on the file, I think this is ready to merge now

PR #436, opened 2017-10-31
FFTCorr

### LINKED ISSUE #434: Bring back the FFT based correlation functions.

rainwoodman:
This PR adds FFT based Correlation Function; closes #434 

Main issue is that the interface between PairCount and FFTPower is different. Not sure which to follow

- FFTPower uses .power, ['power']
- PairCount uses .result, ['xi'].

I also added flags to LinearMesh and LogNormalCatalog to generate unitary / inverted phase realizations.


rainwoodman:
OK!

Issue #405, opened 2017-09-21
Improve CF Estimators vs Pair Counters 


nickhand:

rainwoodman:
At least the documentation shall mention how to construct a Landy-Slazay estimator. What about the other forms of estimators?

nickhand:
I am thinking of a simple wrapper algorithm that takes in an FKPCatalog and computes all of the necessary correlations for the user-specified estimator

rainwoodman:
What about correlation in a PeriodicBox?

nickhand:
we use the analytic randoms calculation in the periodic box to compute the CF in the result already

nickhand:
We need to be clearer about pair counting vs correlation function estimators, which estimators are available, and when we can use analytic vs non-analytic randoms. 

I'd like to hide as much of the survey data vs simulation data as possible, but need to think about how to best do that. It would be nice to provide an equal treatment of angular correlation functions, projected correlation functions, and 3D correlations functions for both simulations and data. 

This will likely overhaul some of the existing stuff we have, so better to do it sooner rather than later. Any thoughts, @rainwoodman?

rainwoodman:
The only input I can think of quickly is that there are two types of scatter point data -- samples of a field (value and weight) and realizations of a density (weight). They can be cross correlated as well. The optimal estimator differs depends on who is correlating with whom. That was why in kdcount I opted to give only sum1 and sum2, and leaving the estimators out. 

Maybe a starting point is to mock up the function signatures a few 'algorithms' that directly does the optimal estimators, then take a look if there is any common ground between them?

nickhand:
I was thinking of only supporting the latter  to start. Is there a common use case in the literature for samples of a field? What are the estimator differences?

Even within just pair counters, we should add better support for correlation function as a function of r and r,mu, projected correlation functions, and angular functions. For simulation cubes we can use analytic randoms, and use one of the estimators for survey data, starting with Landy Slazay but also the less common ones, ie DD/RR - 1

rainwoodman:
Lyman-alpha x quasar was a common use; but maybe nowadays people do it with a faster algorithm (e.g. by los decomposition and looking at the hybrid of power and correlation function?)

Marked correlation function is similar with weight and values. The angular people do it a lot.

There are two ways of constructing this 
- A "correlation function" that takes the type of estimator, data and random.
- Or you have a correlation.LandySlazayEstimator that takes data, and random, and calls the internal paircouting routines.

[...]

PR #439, opened 2017-11-04
WIP: new pair counters + correlation function estimators

nickhand:
Addresses #405.

The goal here is to:

- better wrapping of Corrfunc, including all of the available pair counters
- add functionality for 1d, 2d, projected, and angular pair counters and correlation functions
- add functionality to estimate correlation functions, either using analytic or catalog randoms


nbodykit/algorithms/pair_counters/domain.py
```diff
@@ -0,0 +1,293 @@
...
+        boxsize = boxsize1
+    else:
+        boxsizes = numpy.vstack([boxsize1, boxsize2])
+        argmax = numpy.argmax(boxsizes, axis=0)
+        boxsize = boxsizes[argmax, [0,1,2]]
+
+    # initialize the domain
+    # NOTE: over-decompose by factor of 2 to trigger load balancing
+    grid = [
+        numpy.linspace(0, boxsize[0], 2*np[0] + 1, endpoint=True),
```
rainwoodman: Consider making factor of 2 a parameter.

nbodykit/algorithms/pair_counters/simbox.py
```diff
@@ -0,0 +1,199 @@
...
+    -----
+    This class can compute pair counts using several different coordinate
+    choices, based on the value of the input argument ``mode``. The choices
+    are:
+
+    * ``mode='1d'`` : compute pairs as a function of the 3D separation :math:`r`
+    * ``mode='2d'`` : compute pairs as a function of the 3D separation :math:`r`
+      and the cosine of the angle to the line-of-sight, :math:`\mu`
+    * ``mode='projected'`` : compute pairs as a function of distance perpendicular
+      and parallel to the line-of-sight, :math:`r_p` and :math:`\pi`
```
rainwoodman: Is pi a variable here?

nbodykit/algorithms/pair_counters/corrfunc/base.py
```diff
@@ -0,0 +1,193 @@
...
+        idx = data['npairs'] > 0.
+        for col in self.columns:
+            if col == 'npairs': continue
+            data[col][idx] /= data['npairs'][idx]
+
+        return CorrfuncResult(data)
+
+class MPICorrfuncCallable(object):
+    """
+    A base class to represent MPI-enabled Corrfunc callable.
```
rainwoodman: Perhaps a link or more text what these callables are used for.

nbodykit/algorithms/pair_counters/mocksurvey.py
```diff
@@ -0,0 +1,175 @@
...
+        compute pair counts as a function of the specified coordinate basis;
+        see the Notes section below for specifics
+    first : CatalogSource
+        the first source of particles, providing the 'Position' column
+    edges : array_like
+        the separation bin edges along the first coordinate dimension;
+        depending on ``mode``, the options are :math:`r`, :math:`r_p`, or
+        :math:`\theta`. Expected units for distances are :math:`\mathrm{Mpc}/h`
+        and degrees for angles. Length of nbins+1
+    cosmo : :class:`~nbodykit.cosmology.cosmology.Cosmology`, optional
```
rainwoodman: What if the fiducial cosmology doesn't have radiation?

nbodykit/algorithms/pair_counters/simbox.py
rainwoodman: No it isn't. r_\pi?

rainwoodman:
There is no longer 'xi' any more?

nbodykit/algorithms/pair_counters/simbox.py
nickhand: yes its r_\pi...seems like corrfunc and halotools refer to it as \pi

nbodykit/algorithms/pair_counters/corrfunc/base.py
nickhand: sure no problem

nbodykit/algorithms/pair_counters/mocksurvey.py
nickhand: Do we assume that somewhere? It is only used to compute comoving distance to pass to Corrfunc

nickhand:
I've just added a first pass at the pair counting based 2PCF classes. These classes roughly follow the following logic for xi(r), xi(r,mu), xi(rp, pi), and xi(theta)

1. check if simulation box with PBC turned on; if no randoms catalog explicitly supplied, use analytic randoms and do DD/RR-1
2. otherwise, require a randoms catalog, and do Landy-Szalay estimator to get CF and pair counts

nbodykit/algorithms/pair_counters/corrfunc/base.py
nickhand: One of the things that these classes do are try to handle exceptions from Corrfunc sensibly, since it's a bit broken at the moment (as you've noted). We capture stdout/stderr and the C-level logging, and if the Corrfunc calls, print it all to screen. 

There seems to be helpful C-level logging for errors in Corrfunc, but the Python exception raised is not helpful.

nbodykit/algorithms/pair_counters/mocksurvey.py
rainwoodman: Class can't do without radiation..

nickhand: oh sure, but that's a more general issue with the Cosmology class. We should raise a more helpful message when Tcmb0 is 0 in Cosmology?

nickhand:
I think this is also ready to merge on my end now, if you want to take a look @rainwoodman

nickhand:
@rainwoodman I'm going to merge this because I need it for benchmarks, but we can add updates later if you have comments

PR #440, opened 2017-11-07
Add default columns and do not save them to disk

### LINKED ISSUE #431: allow some columns to not be saved to disk / cleanup hard columns vs defaults
moving the discussion from #427 here. We would like to be able to save certain columns to disk, i.e., do not save default columns to disk. 

This will likely involve a clean up of the "hardcolumn" syntax

nickhand:
This fixes #431.

* introduces "default" columns --> these are columns that are defined via class member functions
* add a metaclass to CatalogSource that picks up default columns and hard coded columns that are part of the class
* add ``is_default`` attribute to ColumnAccessor, letting users test if a column is default; I thought this was better than a ``defaults`` list, which would change as columns are overriden 
* filter out and do not save defaults in CatalogSource.save()

### DIFF
--- a/nbodykit/base/catalog.py
+++ b/nbodykit/base/catalog.py
@@ -11,7 +11,7 @@ import dask.array as da
 
 class ColumnAccessor(da.Array):
     """
-    Provides access to a Column from a Catalog
+    Provides access to a Column from a Catalog.
 
     This is a thin subclass of :class:`dask.array.Array` to
     provide a reference to the catalog object,
@@ -22,8 +22,17 @@ class ColumnAccessor(da.Array):
     that is not explicitly in-place will return
     a :class:`dask.array.Array`, and losing the pointer to
     the original catalog and the meta data attrs.
+
+    Parameters
+    ----------
+    catalog : CatalogSource
+        the catalog from which the column was accessed
+    daskarray : dask.array.Array
+        the column in dask array form
+    is_default : bool, optional
+        whether this column is a default column
     """
-    def __new__(cls, catalog, daskarray):
+    def __new__(cls, catalog, daskarray, is_default=False):
         self = da.Array.__new__(ColumnAccessor,
                 daskarray.dask,
                 daskarray.name,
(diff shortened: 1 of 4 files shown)


nbodykit/base/catalog.py
```diff
@@ -20,15 +22,25 @@ class ColumnAccessor(da.Array):
...
     the original catalog and the meta data attrs.
+
+    Parameters
+    ----------
+    catalog : CatalogSource
+        the catalog from which the column was accessed
+    daskarray : dask.array.Array
+        the column in dask array form
+    is_default : bool, optional
+        whether this column is a default column
```
rainwoodman: Add a comment (thus shall not be serialized)

nbodykit/base/catalog.py
```diff
@@ -65,70 +77,72 @@ def __str__(self):
...
+
+    This can be used as a decorator with or without arguments. If no ``name``
+    is specified, the function name is used.
+
+    Parameters
+    ----------
+    name : str, optional
+        the name of the column; if not provided, the name of the function
+        being decorated is used
+    default : bool, optional
```
rainwoodman: I wonder if the name 'fallback' is better suited than default in this context.

nbodykit/base/catalog.py
```diff
@@ -65,70 +77,72 @@ def __str__(self):
...
+        whether the column is a default column
     """
-    def decorator(getter):
-        getter.column_name = name
+    def decorator(getter, name=name):
+        getter.column_name = getter.__name__ if name is None else name
+        getter.default_column = default
```
rainwoodman: is_default_column?

nbodykit/base/catalog.py
nickhand: I went with "is_default" here and below to unify the syntax. I think we are now using "is_default" everywhere so the syntax reads "column.is_default"

nickhand:
I think this is ready to merge, @rainwoodman ?

rainwoodman:
I am OK with this too.

PR #441, opened 2017-11-08
add a cached halotools catalogs + HOD cleanup

nickhand:
A few additions: 

- HalotoolsCachedCatalog can download + load the built-in halo sims available through Halotools. They seem to be a good size for tutorials (~100 MB) and are useful for comparing 2-pt correlation function results to halotools results
- Support for astropy Table in ArrayCatalog
- re-organize HODCatalog a bit so it subclasses HalotoolsMockCatalog, with can populate a mock in parallel given a Halotools catalog + model. This adds support for generic Halotools models. 
- catch and re-raise exceptions when populating mocks with only the root rank. This should prevent the code from hanging and Halotools is pretty good about having useful exception messages.


nickhand:
@rainwoodman I am ready to merge this now, if you are.

nbodykit/source/catalog/tests/test_hod.py
```diff
@@ -1,175 +1,88 @@
...
-    # test converting from astropy to nbodykit cosmology
-    cat = halos.to_halotools()
-    cat.cosmology = cat.cosmology.to_astropy()
-
-    # make the HOD
-    hod = HODCatalog(cat, seed=42)
-
-    # missing required column
-    del cat.halo_table['halo_id']
+    # no galaxies populated into halos
```
rainwoodman: Because?

nbodykit/source/catalog/tests/test_hod.py
```diff
@@ -1,175 +1,88 @@
...
 
-    # make the HOD catalog from halotools catalog
-    hod = HODCatalog(halos.to_halotools(), seed=42)
+    # populate the mock first
+    hod = halos.populate(Zheng07Model, seed=42)
     size = hod.csize
 
     # repopulate (with same seed --> same catalog)
-    hod.repopulate(seed=42)
+    hod = halos.repopulate(seed=42)
```
rainwoodman: do you like hod.repopulate() better than halos.repopulate()?

rainwoodman: If that's the case then the first ValueError can be eliminated -- without doing populate first there is no hod object for you to repopulate. 

The implication is that the compiled model shall be associated with hod, not halos.

nbodykit/source/catalog/tests/test_hod.py
```diff
@@ -1,175 +1,88 @@
...
     assert hod.csize == size
 
     # repopulate with random seed --> different catalog
-    hod.repopulate()
+    hod = halos.repopulate()
```
rainwoodman: Probably give an explicit seed ? It is unlikely but what if random number generator just gave the same number as you used?

nbodykit/tutorials/tests/test_wget.py
```diff
@@ -0,0 +1,57 @@
...
+    assert filename in os.listdir('.')
+    assert len(os.listdir(filename)) > 0
+
+    # remove the downloaded file
+    shutil.rmtree(filename)
+
+@MPITest([1])
+def test_download_file(comm):
+
+    # download Gadget1P snapshots directory
```
rainwoodman: typo: TPM, not Gadget1P

rainwoodman:
I noted a few minor glitches in the comments. 

hod.repopulate() and halo.populate() can the assertion no repopluate before populate. It encourages moving the compiled model to hod -- but I suspect this works against the data model of halotools. We can still mark halo._repopulate() private and only expose it via hod.repopulate(), then comment about the reason for using a delegation .

nbodykit/source/catalog/tests/test_hod.py
nickhand: logMmin is very large. I'll add a note

nbodykit/source/catalog/tests/test_hod.py
nickhand: I am actually testing that the seed is set by root and then propagated when it is None − I'll update the assertion

nickhand:
Okay, I agree with your repopulate() comment. Now, halos.populate returns a PopulatedHaloCatalog (a light subclass of ArrayCatalog) which has the repopulate() method (which does the repopulation in place). 

I'll merge this if you are ready.

rainwoodman:
Yes looks good!

PR #442, opened 2017-11-10
Fix Memory Error when data is larger than chunk size in paint().

rainwoodman:
reported by @zdplayground

See
https://github.com/bccp/nbodykit/commit/4e41aab09c4c7caa702fe7b9a31f1b54c5923f07#commitcomment-25540401

### DIFF
--- a/nbodykit/base/catalogmesh.py
+++ b/nbodykit/base/catalogmesh.py
@@ -360,7 +360,13 @@ class CatalogMesh(CatalogSource, MeshSource):
 
                 # be sure to use the source to compute
                 position, weight, value = \
-                    self.base.compute(Position[s][sel], Weight[s][sel], Value[s][sel])
+                    self.base.compute(Position[s], Weight[s], Value[s])
+
+                # FIXME: investigate if move selection before compute
+                # speeds up IO.
+                position = position[sel]
+                weight = weight[sel]
+                value = value[sel]
             else:
                 # workaround a potential dask issue on empty dask arrays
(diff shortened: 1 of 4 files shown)


rainwoodman:
@zdplayground : 

You can test this after running:
```
 pip install -U --no-deps https://github.com/rainwoodman/nbodykit/archive/large-data.zip
```
In your conda environment of nbodykit; then restart the kernel.

zdplayground:
@rainwoodman:
After installing large-data.zip, the memory error disappears, though the running time of FFTPower for such case seems a little bit long.

nbodykit/base/catalogmesh.py
```diff
@@ -340,30 +341,26 @@ def to_real_field(self, out=None, normalize=True):
...
 
             if len(Position) != 0:
 
+                # selection has to be computed many times when data is `large`.
+                sel = self.base.compute(Selection[s])
+
                 # be sure to use the source to compute
                 position, weight, value = \
-                    self.base.compute(Position[s], Weight[s], Value[s])
+                    self.base.compute(Position[s][sel], Weight[s][sel], Value[s][sel])
```
nickhand: I think we might want to apply `` sel`` to the arrays after computing to be safer (this was the logic used in 4e41aab0).

rainwoodman:
@zdplayground  it's 1 billion particles on a single core; that's why I usually run this from the job queue with ~ 100 cores.

rainwoodman:
@nickhand let's wait till travis is green before merging.

PR #443, opened 2017-11-10
Add commonly used filter functions.

rainwoodman:
This introduces a new module, nbodykit.filter.

### DIFF
--- a/nbodykit/filter.py
+++ b/nbodykit/filter.py
@@ -1,0 +1,50 @@
+from nbodykit.base.mesh import MeshFilter
+
+import numpy
+
+class TopHat(MeshFilter):
+    kind = 'wavenumber'
+    mode = 'complex'
+
+    def __init__(self, r):
+        """
+            Parameters
+            ----------
+            r : float
+                radius of the TopHat filter
+        """
+        self.r = r
+
+    def filter(self, k, v):
+        r = self.r
+        k = sum(ki ** 2 for ki in k) ** 0.5
+        kr = k * r
+        w = 3 * (numpy.sin(kr) / kr **3 - numpy.cos(kr) / kr ** 2)
+        w[k == 0] = 1.0
+        return w * v
+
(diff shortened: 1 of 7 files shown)


nbodykit/filter.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def filter(self, k, v):
+        r = self.r
+        k = sum(ki ** 2 for ki in k) ** 0.5
+        kr = k * r
+        w = 3 * (numpy.sin(kr) / kr **3 - numpy.cos(kr) / kr ** 2)
+        w[k == 0] = 1.0
+        return w * v
+
+class Gaussian(MeshFilter):
```
nickhand: FourierSpaceGaussian?

nbodykit/filter.py
```diff
@@ -0,0 +1,50 @@
+from nbodykit.base.mesh import MeshFilter
+
+import numpy
+
+class TopHat(MeshFilter):
```
nickhand: FourierSpaceTopHat? KSpaceTopHat? ...we should probably differentiate from the real-space equivalent

rainwoodman:
oops Python 2.x

nbodykit/tests/test_filter.py
```diff
@@ -0,0 +1,37 @@
+from nbodykit.lab import *
+from nbodykit import setup_logging
+from runtests.mpi import MPITest
+
+import pytest
+
+# debug logging
+setup_logging("debug")
+
+from nbodykit.filter import TopHat, Gaussian
```
nickhand: Shall we import filter in nbodykit.lab?

nbodykit/filter.py
rainwoodman: Not sure yet. Probably filter.fourier.TopHat?

rainwoodman: We can always convert the class argument to an argument to support configuration space convolutions.

Issue #402, opened 2017-09-19
global cache?


nickhand:
The cache object is currently owned by the CatalogSource, and repeated calls to compute() need to be occur on the same source to take advantage of the cache. This can present problems when creating new Catalogs (which have new Cache objects) via slicing, etc, or e.g., an FKPCatalog from other sources, b/c the cache is effectively lost. 

We could avoid this with a global Cache, such that any and all calls to compute use the same Cache. Thoughts on this design, @rainwoodman ?

rainwoodman:
Does the cache only catch the 'hard' columns? If a view will eventually go back to the very original hard columns, then this probably isn't a bad issue, assuming only 'hard' columns are slow to load.

rainwoodman:
A global cache will require ways to manage it (context manager?), and a policy governing which data goes into the cache. In contrast, an object bound cache dies when the object dies, so it is a lot simpler.

rainwoodman:
Usually the cache is inserted at the IO layer -- would that be a possibility?

nickhand:
The cache is storing intermediate task operations, so it's a little hard to tell but it seems to end up always caching the data reading step. You can control the size of a given cache (currently set to 1 GB), but this is per cache, so it could actually be easier to control the cache size with a global cache. 

I worry that currently we run the risk of caching the same data more than once (in caches belonging to different objects), needlessly taking up (possibly too much) RAM. The choices regarding which steps stay cached, etc are all performed by the ``cachey``  package, which seems to do a pretty good job. 

Things only go in to the cache if the compute() is executed within the context of the cache, so you do:

```
with GlobalCache:
     arr = dask.compute(...)
```

Alternatively, we can just be a bit more careful about which compute() gets used in algorithms as such. That gets complex pretty quickly with meshes vs catalogs, and was having a hard to ensuring that the same cache was being used during ConvolvedFFTPower

rainwoodman:
With the selection optimization removed this is probably getting more important  ..

nickhand:
fixed via #445

PR #445, opened 2017-11-15
add a global cache

nickhand:
addresses #402 

- GlobalCache class that is resizable
- default size is 100 MB


nickhand:
fixes #402

PR #447, opened 2017-12-07
Add support to PCS resampling windows.

rainwoodman:
This requires version 0.1.37 of pmesh.


rainwoodman:
What do you mean?

nickhand:
Change the behavior so when you set ``compensated=True``, it uses CompensateCIC (TSC,PCS), etc
to do the window de-convolution (vs the current behavior which uses CompensateCICAliasing)

nickhand:
I've pinned our conda build version to <3.1 so we can build the new pmesh and opened an issue about the error we are seeing

rainwoodman:
Why do we want to change the behavior? The current compensation makes sure that the shotnoise spectrum is flat. If you change it it's no longer flat I suppose.

PR #450, opened 2017-12-21
Add an antialiasing mesh source.

rainwoodman:
This shall gradually deprecate the catalog mesh interface.

It currently implements the INTERLACE and GRID scheme, but
new ones can be added.

GRID is slightly better than INTERLACE at the same window;
but it is a lot more expensive.


nbodykit/source/mesh/aamesh.py
```diff
@@ -0,0 +1,392 @@
...
+from nbodykit import _global_options
+import numpy
+import logging
+import warnings
+
+# for converting from particle to mesh
+from pmesh import window
+from pmesh.pm import RealField, ComplexField
+
+class SSAAMesh(MeshSource):
```
rainwoodman: I dislike the name.

PR #458, opened 2018-03-07
Fix Hermitian weights

eelregit:
so that the Nyquist plane (perp. to the hermitian symmetric axis) also has singular weight.
This is consistent with [this](https://github.com/bccp/abopt/blob/049c9531b4325fa71222a79f67b2584400fcc8bf/abopt/vmad3/lib/fastpm.py#L84).

### DIFF
--- a/nbodykit/meshtools.py
+++ b/nbodykit/meshtools.py
@@ -169,7 +169,7 @@ class MeshSlab(object):
 
                 # get the indices that have positive and non-Nyquist freq along symmetry axis
                 symmaxis_ind = numpy.arange(symmaxis_shape)
-                nonsingular = numpy.logical_and(symmaxis_ind > 0, 2 * symmaxis_ind != symmaxis_shape)
+                nonsingular = (symmaxis_ind > 0) & (2 * symmaxis_ind != symmaxis_shape)
                 nonsingular.shape = self._coords[self.symmetry_axis].shape
                 nonsingular = numpy.take(nonsingular, 0, axis=self.axis)
 
--- a/nbodykit/meshtools.py
+++ b/nbodykit/meshtools.py
@@ -138,7 +138,7 @@ class MeshSlab(object):
     @property
     def nonsingular(self):
         """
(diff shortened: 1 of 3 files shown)


eelregit:
Not sure yet how to change the test to reflect this.

nbodykit/meshtools.py
```diff
@@ -155,18 +155,22 @@ def nonsingular(self):
...
                     idx = numpy.zeros(self.shape, dtype=bool)
 
             # one of slab dimensions is symmetry axis
             else:
 
-                # get the indices that have positive freq along symmetry axis
-                nonsingular = (self._coords[self.symmetry_axis] > 0.)
+                # get the indices that have positive and non-Nyquist freq along symmetry axis
+                symmaxis_ind = numpy.arange(symmaxis_shape)
+                nonsingular = numpy.logical_and(symmaxis_ind > 0, 2 * symmaxis_ind != symmaxis_shape)
```
rainwoodman: why not '&'

rainwoodman:
You need to modify the test case in `nbodykit/tests/test_meshtools.py` to reflect the updated logic. The old assertion was that Nyquist weight shall be 2.0 too.

nbodykit/tests/test_meshtools.py
```diff
@@ -75,18 +75,20 @@ def test_hermitian_weights(comm):
     data = data[:] + 1j*data[:]
 
     cfield[...] = data[:]
-    x = cfield.x
+    k = cfield.k
```
eelregit: ComplexField has no k? @rainwoodman

eelregit:
Now I understand everything was correct, which however relies on the convention that the Nyquist frequency is negative. A tiny improvement to relieve such an assumption seems difficult, because it's impossible to know if a mode is on the Nyquist plane without knowing the shape along the symmetric axis before the Hermitian compression.

rainwoodman:
This is sad. But could you change this PR to adding some comments related to this in the source code?

PR #459, opened 2018-03-07
Fix docstrings

eelregit:
See [#458](https://github.com/bccp/nbodykit/pull/458).


rainwoodman:
We shall assert the nyquist is negative in pmesh in one of the test cases.

PR #466, opened 2018-03-14
Split perturbation to MatterDominated and RadiationDominated.

rainwoodman:
MatterDominated doesn't take a cosmology object as input as it
is not a realistic 'cosmology'. This is to emphasis a
drastic approximation must be applied when MatterDominated perturbation
is used.

The old PerturbationGrowth object is now an alias to `RadiationDominated`. Documentation makes it clear this has some approximations under the hood as well.


rainwoodman:
The purpose of this PR is to support pure matter simulations with fastpm. Currently we need to clone to a less radiated cosmology to mimic a nbody simulation with non-radiated background.

PR #478, opened 2018-04-19
WIP: standard reconstruction

rainwoodman:
This PR implements a standard reconstruction algorithm on a periodic box.


rainwoodman:
@martinjameswhite 

Could you recommend some test cases?:

Currently the test case (test_recon.py) I run this on a LogNormalCatalog and assert the large scale power hasn't changed. This is not powerful enough. e.g. If I have the sign of the displacement wrong, the large scale power still won't change.

martinjameswhite:
If you have a halo catalog from either a FastPM run or one of the RunPB boxes then reconstruction should lead to a visible sharpening of the BAO peak.
The code needs to assume a bias and a value of "f" or "beta=f/b" when doing reconstruction, but in general those values don't have to be super-accurate for reconstruction to work pretty well so even a "reasonable guess" should be ok.

martinjameswhite:
In your kernel function you need to "take out" the RSD.  What you have implemented is the real-space version.  Perhaps we want a boolean argument saying whether this is real or redshift space?  Alternatively you can just set beta or f to zero for real-space reconstruction.
So you need to have your halo density field (which generates the displacements) and divide it by b(1+beta mu_k^2) where "b" and "beta" (or "b" and "f") are input parameters to the algorithm [in addition to the smoothing].  Then you multiply the line-of-sight component of the "shift" by 1+f, i.e. you shift more in the redshift direction than the transverse directions.  There are two conventions: one is to shift both the "data" and the "randoms" in the same way.  The other convention is to shift the data as above but when shifting the randoms you don't include the 1+f factor in the line-of-sight direction.  The most common convention is the latter one.

rainwoodman:
I suspect don't have the luxury of running a reasonable resolution simulation in the unit tests. (even the current 128**3 lognormal seems to be killing the testbots) So we'll have to defer a further look to a notebook. I'll take a look at the RSD soon.

rainwoodman:
Reconstruction for dark matter at z=0.

image

rainwoodman:
@martinjameswhite Does this look reasonable?

rainwoodman:
While at this, I also implemented a few other Lagrangian reconstruction schemes.

image

image

It appears the LRR scheme doesn't do much to the BAO, and LF2 is slightly worse than LGS because it is suppressed by 3/7.

@marcel-sl  -- could you take look if this result and implementation look reasonable? The schemes are implemented in `_paint()` function.

rainwoodman:
For a sample with bias ~ 1.0
image


image

PR #479, opened 2018-04-19
Add root=Ellipsis, for (All)GatherArray.

rainwoodman:


rainwoodman:
This function is useful for gathering a distributed column back. Say one want to populate the redshift of randoms following the distribution of data.

PR #480, opened 2018-04-30
bigfile shall look for header starting from root.

rainwoodman:
As specified in the documentation. The rational is that
this way one can always easily specify a special header block in not
in the dataset. If it were relative to the dataset then there is
no way to load the header if it is outside.

This fix solves a few occasions where scripts mysteriously stopped
working when loading catalog.

### DIFF
--- a/nbodykit/io/bigfile.py
+++ b/nbodykit/io/bigfile.py
@@ -85,11 +85,13 @@ class BigFile(FileType):
         """ Find header from the file block by default. """
         if header is Automatic:
             for header in ['Header', 'header', '.']:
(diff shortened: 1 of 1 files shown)


rainwoodman:
I suspect the HDF5 catalog source or the fits source doesn't follow this convention yet.

rainwoodman:
@modichirag @emassara This is why some of your scripts stopped working. the line that forced header discovery to relative to 'dataset' part broke many scripts that reads FastPM catalogs, where 'Header/' and '1/' are at the same level.

PR #486, opened 2018-05-04
ENH: Allow over decomposition in FOF.

rainwoodman:
This feature is useful if we want to run FOF on survey geometry data,
e.g. a lightcone.


rainwoodman:
Travis error may be related to pmesh. Test after new version is tagged.

PR #487, opened 2018-05-08
Linear theory correspondant of a nbody simulation.

rainwoodman:
This PR adds the linear theory correspondant of a nbody simulation.

The initial condition can be set from CLASS transfer function.

The equations are symplectic. It can also be used to construct back-scaled initial conditions, when ran in inverse with high precision.


rainwoodman:
This is intended to be the solution to the boltzmann solver mentioned in https://github.com/MP-Gadget/MP-Gadget/pull/202 .

PR #490, opened 2018-05-15
use saner column / attr names in paircount result.

rainwoodman:
wnpairs : weighted npairs (per bin implied)

total : total (no longer per bin)

paircount.pairs.attrs['total_wnpairs'] shall be prefered than
paircount.attrs['total_wnpairs'] -- leave this change to the next
PR.


nickhand:
@rainwoodman I hope to take a look at this this weekend.

rainwoodman:
Thanks!

PR #497, opened 2018-05-31
Add a take method to BinnedStatistic.

rainwoodman:
The take method behaves more or less like the traditional numpy indexing.
Either a boolean mask of the same shape of data, or boolean masks along
each axis is accept. It is more powerful and sometimes allows more
intuitive code segments than the current sel method; see the example in
docstring.

The full boolean mask is projected along each axis then 'and'ed to the
given mask along the axis.

If a list of indices is given along any axis, it is first converted to
a mask.

In the end the booleans are converted to lists because the finalize method
takes only indices.

### DIFF
--- a/nbodykit/binned_statistic.py
+++ b/nbodykit/binned_statistic.py
@@ -254,7 +254,10 @@ class BinnedStatistic(object):
         edges = {}
         coords = {}
         for i, dim in enumerate(self.dims):
-            idx = indices[i] + [indices[i][-1]+1]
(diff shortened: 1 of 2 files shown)


rainwoodman:
One intention of this PR is to allow one to use very fine dk bins to essentially keep only one unique |k| per bin. Then simply keep only those bins that are ever sampled with

```
    r = FFTPower(field, mode='1d', dk=1e-4)
    pkmu = r.power.take(r.power['modes'] > 0)
```

Issue #484, opened 2018-05-04
Renaming MeshSource.paint() to MeshSource.compute().


rainwoodman:
It provides a more uniform appearance of the API. The documentation may need a search/replacement too.

The old .paint() function can be modified to issue a deprecation warning that points to the new compute() method.

rainwoodman:
The `run()` methods in Algorithms can also be renamed to `compute()`.

rainwoodman:
@nickhand do you recall why we didn't use `compute()` for all of these names?

nickhand:
I don’t think we had a good reason....but I like the more uniform approach

rainwoodman:
Perhaps it is because MeshSource.compute() need an extra argument to decide if real or complex is desired.

PR #498, opened 2018-05-31
Add compute method and remove Catalog interface from CatalogMesh.

rainwoodman:
This is required to add a compute method to MeshSource. Because
if CatalogMesh is a CatalogSource then CatalogSource.compute comes
into the way.

I think this PR also sheds some light on how to deal with the temporary
variables created during FKP/convpower. They probably shall be saved
as attributes of the CatalogMesh objects directly, rather than stored
as columns of the underlying sources.

### DIFF
--- a/nbodykit/algorithms/convpower.py
+++ b/nbodykit/algorithms/convpower.py
@@ -144,9 +144,9 @@ class ConvolvedFFTPower(object):
                     use_fkp_weights=False,
                     P0_FKP=None):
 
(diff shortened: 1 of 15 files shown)


rainwoodman:
This PR is one step closer to a clean merge of #450 . 

I stopped adding .compute() method to the algorithms, because it would require lifting the load / save method to a new class, which seems to be an overkill.

rainwoodman:
Replaces #484

PR #501, opened 2018-06-05
WIP: Add an interface for AdvancedProjection of fields.

rainwoodman:
This PR allows the mode argument of FFTPower to be an AdvancedProjection object, which does allows changing the basis of the Power/XI field, with inner product operators.

This is slower than bin counting, but it allows easier extension to other basis where binning is not the right thing to do.

The plan is to implement ProjectFFTPower algorithm with these Projection objects.

I do not quite like the names of things. These concepts are very similar to the 'spaces' concept defined in nifty, so perhaps some ideas can be borrowed from there?


rainwoodman:
@VMBoehm -- any inputs on this?

Issue #515, opened 2018-08-11
Issues in Base3PCF


adematti:
Hi,

Willing to implement Slepian et al. 2017 (arXiv: 1709.10150v1)'s anisotropic 3pt correlation function, I started from your code.
I noticed two things:
- In the line threeptcf.py#L101, zeta[l,...] += alm.real, I do not think you should use "l" (ell, degree of the harmonics) to index zeta (the code fails e.g. when asking for even multipoles only)
- In the line threeptcf.py#L111, tree_sec.enum(tree_prim, rmax, process=callback, iprim=iprim), the chunk size is by default limited to 100000 (https://github.com/rainwoodman/kdcount/blob/d827e47d8c255ae2ff8d04baa81b01a211040a7e/kdcount/__init__.py#L30). I do not think it is correct to split secondaries in different chunks, as we are making a product of these (their projection onto the Ylms) in the callback function. So I believe the output is wrong when the number of secondaries exceeds 100000.

Thank you for your code,

Cheers

rainwoodman:
I can answer the second question now. Yes you are. The bunch size is small, but the `process` function will be called multiple times -- when we seen enough number of pairs to fill the buffer for the bunch, the `process` function is called once. This reduces the Python function call overhead and allows us to write vectorized numpy operations for many pairs.

The first question I will have to look into.

rainwoodman:
I agree with you the code looks buggy. l is not the correct index. Also the ordering of the return value of `__call__` of YlmCache is undefined, so it may return results in inconsistent sequences on different ranks and cause the result to be wrong when run in parallel. May also cause problems decoding the results to l values.

rainwoodman:
Do you want to file a PR to fix these? Adding @nickhand to this too.

adematti:
Hi,

Sorry I was away last week (and will not be very available this week either).

2) I get that the process function can be called several times for each primary (till all secondaries are treated), that is a clever way to proceed.
However, I argue that the process function cannot be vectorized along the secondaries here, as you are making the product of alms l. 99 of threeptcf.py
(just like (x1+x2+...)*(y1+y2+...) != x1*y1+x2*y2+... in general).
If I set e.g. tree_sec.enum(tree_prim, rmax, process=callback, iprim=iprim, bunch=100) (instead of bunch=100000) l. 111, then the test test_threeptcf.py::test_sim_threeptcf fails.
[...]

PR #516, opened 2018-08-20
WIP: Fixes a few bugs in three point function paircounting.

rainwoodman:
This is working towards the issues pointed out by @adematti in #515.

The correctness test is expected to fail as the old result data set were probably generated from the wrong algorithm. 

Two bugs were fixed:

1. adding support to arbitrary poles input. Previously it has to be continuous starting from 0.

2. fixing the summation of zeta. zeta is ```alm^2```, thus shall not be summed inside callback function.

### DIFF
--- a/nbodykit/algorithms/threeptcf.py
+++ b/nbodykit/algorithms/threeptcf.py
@@ -32,7 +32,7 @@ class Base3PCF(object):
             self.attrs['BoxSize'] = BoxSize
             self.attrs['periodic'] = periodic
 
-    def _run(self, pos, w, pos_sec, w_sec, boxsize=None):
+    def _run(self, pos, w, pos_sec, w_sec, boxsize=None, bunchsize=10000):
         """
         Internal function to run the 3PCF algorithm on the input data and
         weights.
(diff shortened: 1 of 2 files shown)


nbodykit/algorithms/threeptcf.py
```diff
@@ -108,11 +111,23 @@ def callback(r, i, j, iprim=None):
...
+            tree_sec.enum(tree_prim, rmax, process=callback, iprim=iprim, bunch=bunchsize)
 
             if self.comm.rank == largest_load and iprim % chunk_size == 0:
                 self.logger.info("%d%% done" % (10*iprim//chunk_size))
 
+        # combine alms into zeta;
+        # this cannot be done in the callback because
+        # it is a nonlinear function (outer product) of alm.
+        for (l, m) in alms:
+            alm = alms[(l, m)]
```
adematti: I believe alms should be summed after each primary is treated, i.e. this piece of code should be inside the loop on primaries
for iprim in range(len(pos)): (l. 112), just after l. 114
(and alms should be initialized just before l. 114)

rainwoodman:
All green!

PR #521, opened 2018-09-11
Apply the culling optimization to ColumnAccessor.

rainwoodman:
Since dask 0.19.0 we start to get OOM errors, because
dask stopped applying culling to the default compute method when
optimize_graph is False.

We still do not want optimize_graph because it permutes the order
ing of column and row access in a unfavorable way; but we do
want the culling. So implement the culling in the new
graph protocal and see if it helps.


rainwoodman:
My local test shows the problem is fixed. Will merge after Travis is green.

PR #545, opened 2018-12-30
Fix shotnoise estimation of weighted to_mesh() calls.

rainwoodman:
Before this PR we used a simple formula 1 / nbar, which is incorrect if the data is weighted.
@martinjameswhite suggested an alternative formula:

``` V / (sum w**2 / (sum w)**2) ```

It reverts to ```V/N``` when w is unweighted. This expression is also consistent with the shotnoise combination equation in MultipleSpeciesCatalog.

I strongly suspect this was a regression. But anyhow here is the fix.
A test case is added to make sure it does not break again.

### DIFF
--- a/nbodykit/base/catalog.py
+++ b/nbodykit/base/catalog.py
@@ -781,7 +781,7 @@ class CatalogSourceBase(object):
         if window is not None:
             resampler = window
             import warnings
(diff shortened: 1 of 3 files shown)


rainwoodman:
@modichirag

PR #554, opened 2019-01-07
Write all columns as the same time

rainwoodman:


rainwoodman:
This PR creates complicated IO patterns that tends to crash a tightly sized dask cluster.

PR #556, opened 2019-01-15
Catalog.save may return futures.

rainwoodman:
Catalog.save may return futures when compute=False

The use case is if we know some columns can be computed more efficiently if they are computed together, then we can do the store operation together.


rainwoodman:
It would be great if we can have stalled futures that does not gets scheduled immediately.

Issue #575, opened 2019-03-02
Bug in CatalogMesh.to_real_field(), when particle chunk size is throttled


sjforeman:
I believe there is a bug in CatalogMesh.to_real_field. Consider lines 303-332 of nbodykit.source.mesh.catalog. The second line of the "try" attempts to increase chunksize if it has previously been throttled. However, the placement of this increase can lead to particles that are skipped in the loop, and not assigned to the grid. Specifically, the following sequence can occur:
- chunksize starts at value c1, and the loop attempts to grid particles in slice(i,i+c1)
- a StopIteration is encountered, so chunksize is reduced to c2 and the iteration is restarted
- the loop grids particles in slice(i,i+c2)
- chunksize is increased to value c3, with c2<c3<c1
- i is incremented to i+c3
- the next iteration attempts to grid particles in slice (i+c3,i+2*c3)
- *however*, the particles in slice (i+c2+1,i+c3) have been completely skipped!
(I have recently encountered this situation in some tests involving painting Gadget particle snapshots to a mesh).

This can be fixed if the second line of the "try" is relocated to the end of the loop (in line numbers, moving line 312 to line 333, keeping the command within the while loop).

rainwoodman:
Yes. This is indeed a bug. chunksize adjustment shall always occur after the sliding window is moved. Do you want to file a PR with the fix?

PR #576, opened 2019-03-04
Fix painting bug in CatalogMesh

sjforeman:
Fix bug from issue https://github.com/bccp/nbodykit/issues/575

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ Changelog
 
 0.3.11 (Unreleased)
 ------------------
+* :issue:`575`: Fix bug in painting particles to mesh in CatalogMesh
 
 0.3.10 (2019-02-07)
 ------------------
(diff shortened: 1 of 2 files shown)


sjforeman:
Okay, so this build fails in Python 2.7 and 3.6, but passes in 3.5. Even if I reverse my changes (which are just moving one line of code around in source/mesh/catalog.py) and test in a new conda environment, e.g.
```
conda create -n nbodykit-dev python=2
source activate nbodykit-dev
conda install -c bccp --file requirements.txt --file requirements-extras.txt
pip install -e .
python run-tests.py --with-coverage --html-cov
```
I still get a failure in the same place (`test_cache.py`). This makes me wonder whether the failure isn't due to my change, but some other change that was made since the last build. Any thoughts?

rainwoodman:
You are right it is irrelevant.

This seems to be a bug of cachey, but I am not sure why and how we are only triggering it now.

rainwoodman:
This is reproducible on the nbodykit master branch.

Cachey issue filed at https://github.com/dask/cachey/issues/16

I will merge this PR as is. Thanks!

sjforeman:
Great - thanks for looking into that!

PR #579, opened 2019-04-18
Add explicit optional argument, due to the change of default options in mcfit

eelregit:


rainwoodman:
travis fails. Looks like you shall tag a version and make a release on pypi, or do something like that?

eelregit:
... Why it didn't fail for python 3.5?

rainwoodman:
Its probably failing due to another issue related to sympy which stucks at
an older veraipn on 3.5. So your PR is working.

I will merge both later today and make a release. Thanks!

rainwoodman:
Green and merged. Thanks!

eelregit:
Thanks a lot :)

Issue #580, opened 2019-04-22
three point function fails with sympy 1.4


rainwoodman:
**Describe the bug**
After updating sympy to 1.4, Three Point function crashes with an error.
See

https://travis-ci.org/bccp/nbodykit/jobs/522830443#L1520

**Version**
Head of master branch.

PR #581, opened 2019-04-22
Compatibility with sympy 1.4.

rainwoodman:
The replace method in Sympy no longer recognize string exprs.
Must explicitly call sympify on the expr.

This may be a regression bug on Sympy side, but we do not care too
much about this.


rainwoodman:
@nickhand Let me know if there is a cleaner way of dealing with this. I am not sure importing sympify from sympy.core is proper..

rainwoodman:
Closes #580.

PR #592, opened 2019-06-28
Dask update ndim

rainwoodman:
This PR is supposed to fix incompatibility with recent dask versions as seen in 

https://travis-ci.org/MP-Gadget/MP-Gadget/builds/551092287


rainwoodman:
I think this is why. @sbird

sbird:
Thanks for the quick fix

rainwoodman:
I'll tag a release after fixing the release system(s). Only three months and it's all over the place broken...

PR #604, opened 2019-11-03
Allow any glob pattern in file names.

rainwoodman:
Since we claim a glob is supported, we should not only support
'*'.


martinjameswhite:
Yes, looks good ...

PR #610, opened 2020-01-31
Correct odd multipoles in presence of wide-angle effect

akrolewski:
Fixed a bug in the nbodykit odd multipoles:
--correct the definition of  "hermitian_symmetric" to use the "compressed" attribute of the mesh
--ensure that coordinates are real even when mesh is complex
--changed default mesh type to 'c16.' added explanation for the user that dtype='f8' is faster for even multipoles but yields incorrect odd multipoles in the presence of the wide-angle effect.

### DIFF
--- a/nbodykit/algorithms/convpower/catalog.py
+++ b/nbodykit/algorithms/convpower/catalog.py
@@ -1,5 +1,4 @@
-#from nbodykit.source.catalog.species import MultipleSpeciesCatalog
-from species import MultipleSpeciesCatalog
+from nbodykit.source.catalog.species import MultipleSpeciesCatalog
 from nbodykit.transform import ConstantArray
 
 import numpy
@@ -168,12 +167,12 @@ class FKPCatalog(MultipleSpeciesCatalog):
             the number of cells per box side; if not specified in `attrs`, this
             must be provided
         dtype : str, dtype, optional
-            the data type of the mesh when painting. dtype='f8' assumes
+            the data type of the mesh when painting. dtype='f8' or 'f4' assumes
             Hermitian symmetry of the input field (\delta(x) = 
             \delta^{*}(-x)), and stores it as an N x N x N/2+1 real array.
             This speeds evaluation of even multipoles but yields 
             incorrect odd multipoles in the presence of the wide-angle effect.
-            dtype='c16' stores the field as an N x N x N complex array 
+            dtype='c16' or 'c8' stores the field as an N x N x N complex array 
             to correctly recover the odd multipoles.
         interlaced : bool, optional
             whether to use interlacing to reduce aliasing when painting the
@@ -207,7 +206,7 @@ class FKPCatalog(MultipleSpeciesCatalog):
             deprecated. set nbar in the call to FKPCatalog()
         """
         from .catalogmesh import FKPCatalogMesh
-        
+
         if window is not None:
             import warnings
             resampler = window
--- a/nbodykit/algorithms/convpower/catalog.py
+++ b/nbodykit/algorithms/convpower/catalog.py
@@ -149,8 +149,7 @@ class FKPCatalog(MultipleSpeciesCatalog):
 
         return BoxSize, BoxCenter
 
-    def to_mesh(self, Nmesh=None, BoxSize=None, BoxCenter=None, fast_even_multipoles=False,
-    dtype='f4', interlaced=False,
+    def to_mesh(self, Nmesh=None, BoxSize=None, BoxCenter=None, dtype='c16', interlaced=False,
                 compensated=False, resampler='cic', fkp_weight='FKPWeight',
                 comp_weight='Weight', selection='Selection',
                 position='Position', bbox_from_species=None, window=None, nbar=None):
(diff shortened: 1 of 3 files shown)


nbodykit/algorithms/convpower/catalog.py
```diff
@@ -1,4 +1,5 @@
-from nbodykit.source.catalog.species import MultipleSpeciesCatalog
+#from nbodykit.source.catalog.species import MultipleSpeciesCatalog
+from species import MultipleSpeciesCatalog
```
rainwoodman: Revert this.

nbodykit/algorithms/convpower/catalog.py
```diff
@@ -167,7 +168,13 @@ def to_mesh(self, Nmesh=None, BoxSize=None, BoxCenter=None, dtype='f4', interlac
             the number of cells per box side; if not specified in `attrs`, this
             must be provided
         dtype : str, dtype, optional
-            the data type of the mesh when painting
+            the data type of the mesh when painting. dtype='f8' assumes
```
rainwoodman: "'f8' or 'f4'", and "'c16' or 'c8'"

nbodykit/algorithms/convpower/catalog.py
```diff
@@ -200,7 +207,7 @@ def to_mesh(self, Nmesh=None, BoxSize=None, BoxCenter=None, dtype='f4', interlac
             deprecated. set nbar in the call to FKPCatalog()
         """
         from .catalogmesh import FKPCatalogMesh
-
+
```
rainwoodman: Let's avoid introducing new extra spaces.

nbodykit/algorithms/convpower/fkp.py
```diff
@@ -765,13 +765,13 @@ def _cast_mesh(mesh, Nmesh):
     """
     from .catalog import FKPCatalog
     from .catalogmesh import FKPCatalogMesh
-
+
```
rainwoodman: Extra spaces.

nbodykit/algorithms/fftpower.py
```diff
@@ -295,7 +295,7 @@ def run(self):
             kedges, kcoords = _find_unique_edges(y3d.x, 2 * numpy.pi / y3d.BoxSize, kmax, y3d.pm.comm)
 
         # project on to the desired basis
-        muedges = numpy.linspace(0, 1, self.attrs['Nmu']+1, endpoint=True)
+        muedges = numpy.linspace(-1, 1, self.attrs['Nmu']+1, endpoint=True)
```
rainwoodman: This may break some tests. Did you run runtest.py --single  locally?

nbodykit/algorithms/convpower/catalog.py
akrolewski: Fixed this in latest commit.

nbodykit/algorithms/convpower/catalog.py
akrolewski: Fixed this in latest commit.

nbodykit/algorithms/convpower/catalog.py
akrolewski: Fixed this in latest commit.

nbodykit/algorithms/convpower/fkp.py
akrolewski: Fixed this in latest commit.

nbodykit/algorithms/fftpower.py
akrolewski: I'm having some trouble running runtest.py --single locally (seems to be trouble with MPI). Can I just copy the directory to nersc and run runtest.py there?

rainwoodman:
Thats interesting. What errors did you get? Is it a Mac?

It should work more or less the same on nersc computing nodes with the
bcast environments.

akrolewski:
Yes, it's a mac... on python3, I get an import error from mpi4py:

ImportError:
dlopen(/Users/ALEX/anaconda/envs/py35/lib/python3.5/site-packages/mpi4py/
MPI.cpython-35m-darwin.so, 2): Library not loaded: @rpath/libmpi.1.dylib

  Referenced from:
/Users/ALEX/anaconda/envs/py35/lib/python3.5/site-packages/mpi4py/
MPI.cpython-35m-darwin.so

  Reason: image not found


and on python2 I can get the tests to start but then they fail with an MPI
error:


../testenv/lib/python2.7/site-packages/nbodykit/tests/test_binned_stat.py ..

../testenv/lib/python2.7/site-packages/nbodykit/tests/test_meshtools.py
[SOMEONEs-MacBook-Air.local:13735] *** An error occurred in MPI_Comm_size

[SOMEONEs-MacBook-Air.local:13735] *** on a NULL communicator

[SOMEONEs-MacBook-Air.local:13735] *** Unknown error

[SOMEONEs-MacBook-Air.local:13735] *** MPI_ERRORS_ARE_FATAL: your MPI job
will now abort

rainwoodman:
Looks like a mix up of MPI versions on your system -- but could be the mac
binaries are broken again.

The 3.5 binaries are really old (we stopped building them several months
ago) -- what about 3.7?

rainwoodman:
The travis errors are irrelevant, and I've triggered a rerun. 

I am a bit concerned about the test coverage. Your have spent a lot of work validating the correctness of the result.

I'd like to add a test case; not neccesarliy in this PR. What do you think is a minimal non-trival dataset for testing this? Is it possible to generate such a data set with RandomCatalog or LogNormalCatalog?

akrolewski:
Hi Yu,

The mock would need to have direction-dependent RSD in it (not
plane-parallel to the z-axis of the box).  So one could use
LogNormalCatalog for this but the RSD direction would need to be
somewhat different than here:
https://nbodykit.readthedocs.io/en/latest/cookbook/lognormal-mocks.html

Best,
Alex

rainwoodman:
You meant LOS here shall be the direction vector of displace from a
point-wise observer?

cat['RSDPosition'] = cat['Position'] + cat['VelocityOffset'] * LOS

- Yu

akrolewski:
Yes, that's right.

Alex

akrolewski:
Hi Yu,

Should we go ahead and set up a LogNormalCatalog with non-plane-parallel
RSD? Would be great to get this done!

Best,
Alex

rainwoodman:
Yes. We can also merge this and add the tests in a second PR. Do you want to volunteer to do that?

akrolewski:
Hi Yu,

Sure, I can do the second PR with the tests. Probably will be a couple days
until I get to it, but I'll aim to work on it next week.

Best,
Alex

rainwoodman:
Thanks!
The travis failure was irrelevant. I will merge this portion now.

PR #611, opened 2020-02-06
Update fof.py

eelregit:
Some source like the `Gadget1Catalog` does not have an Nmesh attribute

### DIFF
--- a/nbodykit/algorithms/fof.py
+++ b/nbodykit/algorithms/fof.py
@@ -64,9 +64,10 @@ class FOF(object):
         # linking length relative to mean separation
         if not absolute:
             try:
-                mean_separation = pow(numpy.prod(source.attrs['BoxSize']) / source.csize, 1.0 / len(source.attrs['Nmesh']))
+                ndim = len(source.attrs['Nmesh'])
             except KeyError:
-                mean_separation = pow(numpy.prod(source.attrs['BoxSize']) / source.csize, 1.0 / source['Position'].shape[1])
+                ndim = source['Position'].shape[1]
+            mean_separation = pow(numpy.prod(source.attrs['BoxSize']) / source.csize, 1.0 / ndim)
             linking_length *= mean_separation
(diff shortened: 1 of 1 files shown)


nbodykit/algorithms/fof.py
```diff
@@ -63,7 +63,11 @@ def __init__(self, source, linking_length, nmin, absolute=False, periodic=True,
 
         # linking length relative to mean separation
         if not absolute:
-            mean_separation = pow(numpy.prod(source.attrs['BoxSize']) / source.csize, 1.0 / len(source.attrs['Nmesh']))
+            try:
+                ndim = len(source.attrs['Nmesh'])
+            except KeyError:
+                ndim = source['Position'].shape[1]
```
rainwoodman: What about avoid reraise with:
  if 'Nmesh' not in source.attrs

eelregit: Please look at the new changes.

And what do you mean by to "avoid reraise"? Thanks!

eelregit:
Why not including ndim in the attributes of catalog and mesh sources?

nbodykit/algorithms/fof.py
rainwoodman: Not exactly a reraise -- but if there is an exception in line 69, it would be exception inside exception handler.

rainwoodman:
I think it is because if the catalog doesn't define a Position, then ndim is not defined.

How did you test this?

eelregit:
Is there any example of catalog without 'Position'?

And should we move these ndim inference lines to the source class?

rainwoodman:
Without 'Position': Galaxy catalog with RA and DEC.

We probably shouldn't move the ndim inference lines to the source class: one-liner functions that are sometimes defined sometimes not defined is likely not worthy of the complexity.

I think we may already have a few of these, and they are awkward.

rainwoodman:
Merging. Thanks!

PR #613, opened 2020-02-26
Fixed bug in Gaussian filter in nbodykit/filters.py

svonhausegger:
Fixed the return statement in line 58.  Now the Gaussian filter is applied properly to the mesh `v`.


rainwoodman:
Thanks.

The travis failure is unrelated. Merging.

PR #618, opened 2020-05-28
Fixed bug in zhist.py

akrolewski:
zhist.py fails when bins is a scalar. I think this is because "comm" in lines 63 and 64 should be "self.comm."


rainwoodman:
Yes this looks about right. Thanks!

PR #622, opened 2020-07-07
fix to meshtools to avoid nans in legpoly at k=0

akrolewski:
At kx=ky=kz=0, in slab.mu(los), the norm2 is 0 and thus mu is nan.  This leads to a nan in legpoly which propagates into nans for all multipoles in the bin containing k=0 for multipoles other than the monopole.  following suggestion of @michaelJwilson, modify SlabIterator.mu in meshtools to fix this issue.


rainwoodman:
Thanks!

Issue #627, opened 2020-07-30
mesh size changes when running with mpi


xiaohanzai:
I'm trying to calculate the power spectrum for an N-body sim.  There are too many particles so I have to load in a subset of all the particles at a time in order not to run out of memory.  So what I'm doing is like below:

f_sum = np.zeros((256,256,256))
for i in range(26): # 26 is the number of subsets of particles
    data = ... # load in the i-th subset of the particles
    cat = ArrayCatalog({'Position' : data['Position'])
    mesh = cat.to_mesh(resampler='tsc', Nmesh=256, compensated=True, interlaced=True)
    f = mesh.to_real_field()*len(data['Position'])
    f_sum += f
f_sum /= total # of particles
mesh_sum = ArrayMesh(f_sum)
r = FFTPower(mesh_sum, mode='1d')

Since the above is too slow (each subset has ~1e9 particles), I'm trying to parallelize the code with mpi4py, and the code is like this:

comm = CurrentMPIComm.get()
rank = comm.rank
size = comm.size
for i in range(rank, 26, size):
    same as above...
if rank == 0:
    comm.Reduce(...)

The problem is that when I run the code with more than 1 processors, the shape of the f array also changes, so it's not (256,256,256) anymore.  mpirun -n 1 is still fine, but mpirun -n 2 gives (128, 256, 256) and (128, 256, 256), and -n 3 gives (256, 86, 256), (256, 86, 256), (256, 84, 256).  I'm not sure what's happening here.  What I hope the code would do is that each processor can do the computation for one subset of particles, and in the end I can sum up the results from all processors, but it doesn't seem to work like this?  Maybe I have understandings about mpi4py as well...

rainwoodman:
Indeed. The data model of nbodykit is different from your expectation. The data in nbodykit is always distributed to all ranks in the MPI communicator, each of the MPI ranks more-or-less runs identical computing on its own partition of the data.

Therefore, the data you pass into ArrayCatalog must already be distributed by the MPI ranks.

As your data comes with 26 subsets, I would start with 26 MPI ranks, each creates a local ArrayCatalog from its piece of data. Then you can use to_mesh() to obtain a distributed mesh, which can then be handled by FFTPower, something like this:
```
comm = MPI.COMM_WORLD
assert comm.size == 26  # otherwise we miss data
data = loaddata(file_id = comm.rank)
cat = ArrayCatalog(data ..., comm=comm)
mesh = cat.to_mesh(....)
[...]

PR #628, opened 2020-08-03
FileType + FileCatalog cleanup

rainwoodman:
This PR adds an example using a customized FileType in test_file.py.

Some clean up to make it slightly easier to use FileType.
Adding a FileCatalog class.

### DIFF
--- a/nbodykit/source/catalog/tests/test_file.py
+++ b/nbodykit/source/catalog/tests/test_file.py
@@ -148,3 +148,35 @@ def test_stack_list(comm):
 
     os.unlink(tmpfile1)
     os.unlink(tmpfile2)
+
+
+# MyFileType is Used by test_file_type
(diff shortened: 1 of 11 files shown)


nbodykit/source/catalog/tests/test_file.py
```diff
@@ -148,3 +148,35 @@ def test_stack_list(comm):
 
     os.unlink(tmpfile1)
     os.unlink(tmpfile2)
+
+
+# MyFileType is Used by test_file_type
+from nbodykit.io.base import FileType
+
+class MyFileType(FileType):
```
rainwoodman, line 156: @xiaohanzai

You can also try to follow the example here for adapting the different number of files and mpi ranks. #627 (after installing the nbodykit from this PR).

PR #629, opened 2020-08-05
Fix FFTPower to only use positive mu by default

akrolewski:
As pointed out by Grant Merz, FFTPower now bins mu between -1 and 1.  However, often it will be passed fields that are only defined for positive mu values, and this creates unexpected behavior and nans. This PR changes FFTPower to use positive mu by default (as it did before #610), and to allow the user to request it to use negative mu bins if they desire.

### DIFF
--- a/nbodykit/algorithms/fftpower.py
+++ b/nbodykit/algorithms/fftpower.py
@@ -187,11 +187,15 @@ class FFTPower(FFTBase):
     poles : list of int, optional
         a list of multipole numbers ``ell`` to compute :math:`P_\ell(k)`
         from :math:`P(k,\mu)`
+    use_negative_mu : bool, optional
+        set to True if you want ``mu`` bins to range from -1 to 1
+        else they range from 0 to 1
     """
     logger = logging.getLogger('FFTPower')
 
     def __init__(self, first, mode, Nmesh=None, BoxSize=None, second=None,
-                    los=[0, 0, 1], Nmu=5, dk=None, kmin=0., kmax=None, poles=[]):
+                    los=[0, 0, 1], Nmu=5, dk=None, kmin=0., kmax=None, poles=[],
+                    use_negative_mu=False):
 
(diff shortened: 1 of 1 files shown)


rainwoodman:
I am a bit surprised that this doesn't break any test cases.

Could you or @grantmerz elaborate the unexpected behaviour for record? Is it possible that we can fix it without adding a flag? For example, special case if the type of the field is compressed?

akrolewski:
The following code leads to the unexpected behavior (generating the mesh with ArrayMesh and then using FFTPower to get 2D P(k)).  The mesh is real, so its Fourier transform has no negative mu values, but FFTPower histograms mu between -1 and 1 and thus spits out all nans for the first 5 bins in mu.  The mesh is still real even if I pass dtype='c16' to ArrayMesh.

initname = '/global/cscratch1/sd/g4merz/recon/den0512.bin'
init_dat = np.fromfile(initname,dtype=np.float32,sep='')
arr_init = init_dat.reshape((N,N,N),order='F')
mesh_init = ArrayMesh(arr_init,BoxSize=1000)
r_cross = FFTPower(mesh_init, mode='2d', Nmesh=512, Nmu=10, dk=0.005, los=[0,0,1])

Yes, checking whether the field is compressed should work--I'll look into it.

rainwoodman:
Thanks for the explanation. 

This is probably the compressed attribute to look for:
https://github.com/rainwoodman/pmesh/blob/70da89e729911e6c2ae9e56fb2a4e8d41b34b46f/pmesh/tests/test_pm.py#L274

Looks like when the field is compressed we also could use the symmetry to fill the negative mu bins?

rainwoodman:
I think we agreed use_negative_mu and whether the field is compressed are degenerated. Perhaps add a comment in the docstring that use_negative_mu=True usually shall be paired when the corresponding real field type is c16 or c8, if you still have the branch?

PR #651, opened 2021-03-07
Switch to github actions

rainwoodman:


rainwoodman:
OS-X still doesn't work (need to investigate conda-channel-bccp, and find a osx system), but this is better than nothing works..

PR #654, opened 2021-04-08
DOC fix random uniform call in 1 minute intro

eickenberg:
Updating to what I am guessing was meant.
I can also add an `import numpy` to the import statement above for self-containedness.


eickenberg:
Also `result = FFTPower(mesh, Nmu=5)` seems to require a `mode` keyword, e.g. `mode='1d'`

rainwoodman:
Thanks for the fix! The travis failure is unrelated.

rainwoodman:
This is a code snippet (albeit a big one). I'd say missing an import numpy is somewhat tolerable.

PR #656, opened 2021-05-08
Fix workflow error that's been awhile.

rainwoodman:


rainwoodman:
Finally got this working!

Issue #660, opened 2021-12-13
mocks for a non-gaussian field.


Jayashree-Behera:
Hi, I am working on a project based on non-gaussian fields. So while generating mocks using nbodykit (let's say lognormal mocks), one requires to give the matter power spectrum(as a functional argument ) and bias (as a number) as input (among other parameters). This is fine as long as we are working are in a gaussian field but when it comes to non-gaussian field, bias is no longer an independent number but a function of k, hence a functional argument. I was wondering if this feature could be added where one could generate mocks in non-gaussian fields. I have some working code that I could share if you are interested. :)
Thanks,
Shree

Jayashree-Behera:
Hi, thanks for replying. 
Yes you were right. The way I did it is, write a script to get the galaxy power spectrum in a non-gaussian field. Then one can generate the lognormal mocks (using the current code) by just giving this as an argument (bias can be set to None or 1).  I hope I was clear. I can a pull request and maybe, you can check it out for yourself?

Jayashree-Behera:
Hi.. Thanks for replying. 
My code basically lies in the lines of this "..Or did you mean introducing a scale dependent bias to the linear field (gaussian field), before Log Normal transformation? In that case, the scale dependent biased linear field just need a different powerspectrum or transfer function --..."

I wrote a script for the galaxy power spectrum in a non-gaussian field (which incorporates the scale dependent bias). Then I generated the mocks by giving the galaxy power spectrum as the input parameter setting bias =1.
Let me know if this wasn't clear. 

Have agood day,
Shree

rainwoodman:
Thanks. Aftering thinking about this a bit more, I also realized the
displacement field (for velocity) of LogNormal is unbiased(or I think it
is). If we trivially apply the scale dependent bias to the initial gaussian
field, the velocity would be contaminated. We'll need an additional scale
dependent bias step in just for computing the number density of the tracers.

Can I take a look at your implementation?

Jayashree-Behera:
Sure. Should I make a pull request and you can check it out there?

Jayashree-Behera:
Hi.. Did you get a chance to look at the pull request? Does it make sense?

rainwoodman:
Oops. I missed your PR. I'll take a look this week and get back to you.

Jayashree-Behera:
Hi..Did you get a chance to review the updated version in PR?  I have made some modifications and added the unit test file.

sbird:
[...]

PR #662, opened 2021-12-23
mocks from galaxy power spectrum for a non-gaussian field

Jayashree-Behera:
in response to issue #660


nbodykit/cosmology/power/galaxy.py
```diff
@@ -0,0 +1,128 @@
...
+        the object giving the cosmological parameters
+    sigma8 : float
+        the z=0 amplitude of matter fluctuations
+    redshift : float
+        the redshift to compute the power at
+    transfer : str
+        the type of transfer function used
+    """
+    
+    def __init__(self, cosmo, redshift ,b0 ,fNL ,p ,Omega_m ,H0=73.8 ,c=3e5 ,transfer='CLASS'):
```
rainwoodman, line 50: Why not using cosmo.h and cosmo.Omega_m?

Jayashree-Behera: Oh I completely missed that. That would be so much better. Thanks. Any other recommendations?

nbodykit/cosmology/power/galaxy.py
```diff
@@ -0,0 +1,128 @@
+import numpy
+from . import transfers
+from ..cosmology import Cosmology
+from .linear import LinearPower
+
+class GalaxyPower(object):
```
rainwoodman, line 6: Perhaps rename this to FNLGalaxyPower?

nbodykit/cosmology/power/galaxy.py
```diff
@@ -0,0 +1,128 @@
...
+        converted to the proper type
+    redshift : float
+        the redshift of the power spectrum
+    transfer : str, optional
+        string specifying the transfer function to use; one of
+        'CLASS', 'EisensteinHu', 'NoWiggleEisensteinHu'
+    b0 : float
+        the linear bias of the galaxy in a gaussian field
+    fnl : float
+        the non-gaussian parameter
```
rainwoodman, line 25: "Primordial non-gaussian parameter"?

nbodykit/cosmology/power/galaxy.py
```diff
@@ -0,0 +1,128 @@
...
+    p : float
```
rainwoodman, line 26: Does this parameter have a more descriptive name from the literature? It is like a bias correction parameter (removed from b)? larger p -> lower clustering?

What is a recently merged halo? A halo that recently experienced a merger event from progenitors of similar masses?

nbodykit/cosmology/power/galaxy.py
```diff
@@ -0,0 +1,128 @@
...
+        self.H0=H0
+        self.c=c
+
+        self.transfer = transfer
+
+        # set redshift
+        self.redshift = redshift
+        
+    
+    def corrected_bias(self, k):
```
rainwoodman, line 75: Does the literature call this a "corrected bias"? The return value is called a 'total_bias'. This is a bit confusing...

nbodykit/cosmology/power/__init__.py
```diff
@@ -1,4 +1,5 @@
+from .galaxy import GalaxyPower
```
rainwoodman, line 1: Let's also add a unit test or two that exercises this file,, e.g., in this file: https://github.com/bccp/nbodykit/blob/a387cf429d8cb4a07bb19e3b4325ffdf279a131e/nbodykit/cosmology/tests/test_power.py

nbodykit/cosmology/power/galaxy.py
rainwoodman: Is there a paper that we can reference for the formula used here? Perhaps we can name the class after that paper?

Jayashree-Behera: Hi, here is a paper that has all the formulas and explanations. 

https://arxiv.org/pdf/2106.13725.pdf

nbodykit/cosmology/power/galaxy.py
Jayashree-Behera: Thanks. I have made all those changes and updated the repo. I will add the unit test part soon. Thanks again.

Jayashree-Behera:
Hi. I have made all the changes and added the unit test functions. Please take a look and let me know.

nbodykit/cosmology/power/galaxy.py
Jayashree-Behera: Both are right but yes, it was confusing to use 2 different terms so I only used "total bias". Some works also refer to it as the "non-gaussian bias" or "fnl-bias". I am not sure which one would be ideal to use here.  Let me know your thoughts on this. Thanks.

Jayashree-Behera:
Hi, Did you get a chance to take a look into the pull request?
 Thanks.

PR #700, opened 2025-10-24
cosmo_seed

ochase10:
I added a cosmo_seed parameter to give the option for separating the density realization from the sampling realization. That gives the ability to sample the same initial conditions repeatedly and getting different discrete points. That way, one can build up a large catalog by adding up smaller catalogs without generating the large one all at once.


sbird:
Is it just that you want to sample the HOD mocks separately from the initial density field?

